PR #11, opened 2022-04-03
Add Datafold

alexey-milovidov:


alexey-milovidov:
@rfraposa Please add all the remaining entries from https://news.ycombinator.com/item?id=30878761
and also all that we discussed in Slack.

rfraposa:
Manually added to avoid conflict

PR #48, opened 2022-04-21
feat(config): use svg logo in footer

ernado:
Before:
image

After:
image


rfraposa:
This seems like a good change. I will test it and see how it looks. 

(FWIW: the logo was an .svg at some point, then we changed it to a .png and I can't remember why. Maybe this change will remind me 😄  )

rfraposa (approved):

PR #127, opened 2022-05-19
translate-guides-to-zh

zzsmdfj:
1、I'm not sure how this project will be linked to the official documentation, so I'm not sure the current zh/ directory structure is the right one.

2、

image

I'm a little confused here, as I understand it, there should actually be a Last mark (1083) that holds the maximum primary key. Mark 1082 should is still storing the minimum values for the primary key columns of granule 1082(The diagram is correct, but the text is confusing).


rfraposa:
Except for your typo fixes - those can stay in this repo. I will reopen this PR and you can just remove the /zh files.

zzsmdfj:
> Except for your typo fixes - those can stay in this repo. I will reopen this PR and you can just remove the /zh files.

ok. thanks, then, for the path of the image,  I will adjust to use absolute path.

DanRoscigno (approved):
LGTM, thanks @zzsmdfj

PR #136, opened 2022-05-24
Restructuring into smaller, easier to consume, sub-articles.

tom-clickhouse:


DanRoscigno:
@rfraposa also the css file is changed.

tom-clickhouse:
@DanRoscigno Just pushed a commit with the excalidraw source for all the images.

PR #141, opened 2022-05-27
Update Apache Superset guide with driver versions and remote SQLAlchemy URI

DanRoscigno:
Adds a URI to a remote ClickHouse server
Adds driver versions that work with Superset 1.5.0


rfraposa (approved):

DanRoscigno:
@genzgd I replaced the drivers with `clickhouse-connect`, added info about the ports etc., and updated the images/text to work with the table in play.clickhouse.com.  Can you have a look when you get a chance please?

PR #143, opened 2022-05-31
[bugfix] fix the data insert sql

mufasa007:
the past insert sql is not right for now，for error S3 format 
i fix the sql and had tested success!

### DIFF
--- a/docs/en/tutorial.md
+++ b/docs/en/tutorial.md
@@ -87,8 +87,10 @@ Now that you have a table created, let's add the NYC taxi data. It is in CSV fil
     INSERT INTO trips
         SELECT * FROM s3(
             'https://datasets-documentation.s3.eu-west-3.amazonaws.com/nyc-taxi/trips_{1..2}.gz',
-            'TabSeparatedWithNames'
-        )
+            'TabSeparatedWithNames',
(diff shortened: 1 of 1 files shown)


DanRoscigno:
testing now

DanRoscigno:
@mufasa007 can you tell me the version of ClickHouse server that you are using and send the error message please?  The tutorial works fine as is for me on the current version of ClickHouse.  You can get the version by running `select version()`

mufasa007:
the error version is 21.12.3.32

DanRoscigno:
Hi Mufasa, thanks for your PR, I appreciate that you took the time to give us this fix.  In one of the releases since December (your 21.12 version was released in December) the syntax was simplified so that the schema of TSV files (and CSV) could be inferred and it is no longer necessary to provide the column names.  If it is OK with you I will push some changes to your PR to let everyone know about the improvement.  See [this blog](https://clickhouse.com/blog/whats-new-in-clickhouse-22-1/), or the changelog for the details.

mufasa007:
I had checked it by myself . 
The newest version (22.5.1.2079) of clickhouse is right for the "error" s3 sql .  I got it

PR #156, opened 2022-06-04
Add Ingest dir under Guides and TSV guide

DanRoscigno:
This guide uses the NYPD Complaint dataset (see https://github.com/ClickHouse/ClickHouse/issues/30629 ).  After discussing this with @rfraposa we decided to have an Ingest dir under Guides and populate it with guides that provide more information on why decisions are made.  For example, the process of deciding what types to use for columns is included (the number of distinct strings is fewer than 10,000, so use a LowCardinality(String)).

Notes for reviewers:
- don't hold back.  I had some feedback early on, but my goal is to publish something that people can learn best practices from, not a hello world type of guide.
- Thanks for your help and advice!


alexey-milovidov:
Please also add it to the "example datasets" section.

DanRoscigno:
Will do.

DanRoscigno:
Note: @tom-clickhouse found an issue, in the most recent dataset the first column, CMPLNT_NUM, occasionally has the string `H1` added to the complaint number, which causes the parsing to fail if the column is inferred as a numeric (which it is by default, as the first 1,000 records all have numbers in that field).  To fix this I added the setting `input_format_max_rows_to_read_for_schema_inference=2000` to catch these records.

DanRoscigno:
This is ready for review.  The commands are updated to scan the first 2,000 rows of the TSV to infer the schema, as there are a few complaint numbers that are strings after the first thousand are all integers.  I added some information on the ORDER BY based on @tom-clickhouse 's new doc on [key field ordering](https://clickhouse.com/docs/en/guides/improving-query-performance/sparse-primary-indexes/sparse-primary-indexes-cardinality).  Thanks @Ryado @tom-clickhouse

rfraposa (approved):

PR #160, opened 2022-06-07
Using unique paths in ClickHouse keeper

DanRoscigno:
This brings techniques from the experts in ClickHouse Support Services into the docs.  
1. Adds a subdir for ClickHouse Keeper under User Guides > SRE Guides >
2. Moves the existing ClickHouse Keeper guide into the new subdir
3. Adds a new guide on why using unique paths in ClickHouse Keeper (or ZooKeeper) is important, and how to do it.

closes https://github.com/ClickHouse/internal-knowledge-base/issues/5


DanRoscigno:
> This is fine but it would be nice to understand "how" the UUID actually works magically across nodes. That's the question most people will ask.

Me too!  I need a hand.  I am working through the previous guide and this one now to understand more.  I did just push an update to the opening paragraph that I think helps.

genzgd:
Yeah I don't exactly understand it and I've not had the chance to research it or harass the development team.

Also I think we need to be explicit that this requires an Atomic database engine (which is the current default, but we might be able to avoid some support questions if we explicitly call it out).

DanRoscigno:
> Yeah I don't exactly understand it and I've not had the chance to research it or harass the development team.
> 
> Also I think we need to be explicit that this requires an Atomic database engine (which is the current default, but we might be able to avoid some support questions if we explicitly call it out).

Line 275 calls out the need for Atomic, should I add a prereq section and add it there?

e-mars:
i'm not sure how it does its sync with clickhouse keeper either, i can see the table_name-->uuid in system.tables.
I added the atomic requirement in the `troubleshooting` section, maybe needs to be stated somewhere else also? maybe as a note?

genzgd:
Ah missed that -- probably should be moved up before all the instructions, yes.  It would be nice to know what version of ClickHouse that was introduced in too.  Another thing I don't actually know.

DanRoscigno:
It is just `db_uuid` that needs to be Atomic, right?  I will put it there, OK?

genzgd:
Yes, I think it needs to be CREATE DATABASE db_uuid ENGINE Atomic ON CLUSTER . . .

e-mars:
yes, new version by default, will create `atomic` but just in case users have an old version, doesn't hurt any to put in there explicitly.

DanRoscigno:
I am going to build out a docker env and run through the guide(s).  I added ENGINE Atomic, and will grab fresh output from the commands.

DanRoscigno:
I am happy, I tested the guide and it works.  I updated the create statement for the DB, ENGINE comes after CLUSTER, and I added the output.

PR #162, opened 2022-06-08
Add debug info, fix the issue with merges from forks

Felixoid:


DanRoscigno (approved):
LGTM, thanks for this!

Felixoid:
It's not the time yet. Depends on https://github.com/ClickHouse/ClickHouse/pull/37925

Felixoid:
Now it works as expected and find a few issues!

PR #168, opened 2022-06-13
Update skipping-indexes.md

lzydmxy:
Corrected the original incorrect description of the ngrambf_v1 example


DanRoscigno:
Thanks for the PR @lzydmxy 

Because the line you fixed may span lines I think it looks better in a code block.  I tried adding it in the review, but the markdown broke the github review pane.  Trying here:

* The specialized **ngrambf_v1**. This index functions the same as the token index.  It takes one additional parameter before the Bloom filter settings, the size of the ngrams to index.  An ngram is a character string of length `n` of any characters, so the string `A short string` with an ngram size of 4 would be indexed as:
  ```
  'A sh', ' sho', 'shor', 'hort', 'ort ', 'rt s', 't st', ' str', 'stri', 'trin', 'ring'
  ```
This index can also be useful for text searches, particularly languages without word breaks, such as Chinese.

It now looks like so:
image

Let me know if it looks ok and I will push, or you can keep it the way you have it.

PR #209, opened 2022-07-05
add notes about ACID transactions

DanRoscigno:
I made very few changes, the two that I recall are changing `max_insert_size` to `max_insert_block_size`, and changing the default from 1,000,000 rows to 1,048,576 rows.

### DIFF
--- a/docs/en/guides/developer/transactional.md
+++ b/docs/en/guides/developer/transactional.md
@@ -4,7 +4,7 @@ INSERTs into one partition in one table of the MergeTree\* family up to `max_ins
 
 ### Atomic
 
-INSERTs succeed or are rejected as a whole- if a confirmation is sent to the client, then all rows were inserted; if an error is sent to the client, then no rows were inserted.
+INSERTs succeed or are rejected as a whole - if a confirmation is sent to the client, then all rows were inserted; if an error is sent to the client, then no rows were inserted.
 
 ### Consistent
 
@@ -22,7 +22,7 @@ A successful INSERT is written to the filesystem before sending a confirmation t
 
 - If a table has multiple partitions and the INSERT covers multiple partitions - then insertion into each partition is transactional on its own.
 
-- INSERTs into multiple tables with one statement is possible if materialized views are used.
+- INSERTs into multiple tables with one statement are possible if materialized views are used.
 
 - INSERTs into a distributed table is not transactional as a whole, while insertion into each shard is transactional.
 
@@ -37,14 +37,12 @@ A successful INSERT is written to the filesystem before sending a confirmation t
 
 - All ACID properties are valid even in the case of server kill / crash.
 
-- Using either insert_quorum into different availability zones or `fsync` should be enabled to ensure durable inserts in the typical setup.
+- Using either `insert_quorum` into different availability zones or `fsync_after_insert=1` to ensure durable inserts in the typical setup.
 
-- "Consistency" in ACID terms does not cover the semantics of distributed systems, see https://jepsen.io/consistency which is controlled by different settings ([`select_sequential_consistency`](../../operations/settings/settings.md/#settings-select_sequential_consistency))
+- "Consistency" in ACID terms does not cover the semantics of distributed systems, see [Jepsen](https://jepsen.io/consistency) which is controlled by different settings ([`select_sequential_consistency`](../../operations/settings/settings.md/#settings-select_sequential_consistency))
 
 - This explanation does not cover a new transactions feature that allows full-featured transactions for multiple SELECTS over multiple tables and materialized views, etc.
 
 - If a client does not receive response from the server, the client does not know if the transaction succeeded and it can repeat the transaction, using exactly-once insertion properties.
-:::note
-With `async_insert` deduplication will not work, so using retry will result in at least once transaction.
-:::
+- With `async_insert` deduplication will not work per INSERT, so retries on failures will result in at least once semantics of INSERTs.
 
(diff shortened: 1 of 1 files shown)


docs/en/guides/developer/transactional.md
```diff
@@ -0,0 +1,50 @@
+# Transactional (ACID) support
+
+INSERTs into one partition in one table of the MergeTree\* family up to `max_insert_block_size` rows are transactional (ACID):
+
+### Atomic
+
+INSERTs succeed or are rejected as a whole- if a confirmation is sent to the client, then all rows were inserted; if an error is sent to the client, then no rows were inserted.
```
qoega, line 7:
```suggestion
INSERTs succeed or are rejected as a whole - if a confirmation is sent to the client, then all rows were inserted; if an error is sent to the client, then no rows were inserted.
```

docs/en/guides/developer/transactional.md
```diff
@@ -0,0 +1,50 @@
+# Transactional (ACID) support
+
+INSERTs into one partition in one table of the MergeTree\* family up to `max_insert_block_size` rows are transactional (ACID):
```
qoega, line 3: Are you sure this formatting is enough to understand that next points are valid only if `up to `max_insert_block_size` rows` is true? 
Actually for ACI. D is valid in any case.

docs/en/guides/developer/transactional.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Isolated
+
+Clients observe a consistent snapshot of the table - either the state of the table before an INSERT, or the state after a successful INSERT; no partial state is seen.
+
+### Durable
+
+A successful INSERT is written to the filesystem before sending a confirmation to the client, on single replica or multiple replica systems (this is controlled by the `insert_quorum` setting), and ClickHouse can ask the operating system to sync the filesystem data on the storage media (this is controlled by the `fsync_after_insert` setting).
+
+### Details
```
qoega, line 21: I would not add it on the same level as ACID headers. Probably converting previous headers to a table or just a list will make it less tricky.

docs/en/guides/developer/transactional.md
```diff
@@ -0,0 +1,50 @@
...
+
+- If a table has multiple partitions and the INSERT covers multiple partitions - then insertion into each partition is transactional on its own.
+
+- INSERTs into multiple tables with one statement is possible if materialized views are used.
```
qoega, line 25: Probably something like that is more correct? But I'm not a native speaker.
```suggestion
- Atomic INSERT into multiple tables with one statement is possible if materialized views are used.
```
@alexey-milovidov How do you think, do we need to cover here what happens if MV insert fails and how deduplication works for MV?

```suggestion
- INSERTs into multiple tables with one statement are possible if materialized views are used.
```

docs/en/guides/developer/transactional.md
```diff
@@ -0,0 +1,50 @@
...
+- Atomicity is ensured even if `async_insert` is enabled, but can be disabled with the `wait_for_async_insert` setting.
+
+- The `max_insert_block_size` is 1,048,576 rows by default and can be adjusted as needed.
+
+
+- ClickHouse uses multiversion concurrency control (MVCC) with snapshot isolation internally.
+
+- All ACID properties are valid even in the case of server kill / crash.
+
+- Using either insert_quorum into different availability zones or `fsync` should be enabled to ensure durable inserts in the typical setup.
```
qoega, line 40:
```suggestion
- Using either `insert_quorum` into different availability zones or `fsync_after_insert=1` to ensure durable inserts in the typical setup.
```

docs/en/guides/developer/transactional.md
```diff
@@ -0,0 +1,50 @@
...
+
+- "Consistency" in ACID terms does not cover the semantics of distributed systems, see https://jepsen.io/consistency which is controlled by different settings ([`select_sequential_consistency`](../../operations/settings/settings.md/#settings-select_sequential_consistency))
```
qoega, line 42:
```suggestion
- "Consistency" in ACID terms does not cover the semantics of distributed systems, see [Jepsen](https://jepsen.io/consistency) which is controlled by different settings ([`select_sequential_consistency`](../../operations/settings/settings.md/#settings-select_sequential_consistency))
```

docs/en/guides/developer/transactional.md
```diff
@@ -0,0 +1,50 @@
...
+
+- This explanation does not cover a new transactions feature that allows full-featured transactions for multiple SELECTS over multiple tables and materialized views, etc.
+
+- If a client does not receive response from the server, the client does not know if the transaction succeeded and it can repeat the transaction, using exactly-once insertion properties.
+:::note
+With `async_insert` deduplication will not work, so using retry will result in at least once transaction.
+:::
```
qoega, line 49:
```suggestion
- With `async_insert` deduplication will not work per INSERT, so retries on failures will result in at least once semantics of INSERTs.
```

docs/en/guides/developer/transactional.md
DanRoscigno: Thanks @qoega , I updated, here is how it reads now.  Also moved the ACID lines from H2 to a dictionary:
image

docs/en/guides/developer/transactional.md
DanRoscigno: see above screen-grab

docs/en/guides/developer/transactional.md
alexey-milovidov: Let's merge, then improve...

PR #224, opened 2022-07-15
Fix images background in dark mode.

alexey-milovidov:


src/css/custom.css
```diff
@@ -205,10 +205,9 @@ code {
   border: 2px solid var(--clickhouse-yellow);
 }
 
-/* Add a drop shadow to images in the user guides and docs */
 .image {
   width: 100%;
-  filter: drop-shadow(2px 4px 6px gray);
```
alexey-milovidov, line 211: The shadow was not in effect for some reason, so, I've removed it.

DanRoscigno:
Closing, we are using substring matching instead at least for now.

PR #227, opened 2022-07-17
A minor wording change

mrd000:
There was a mistake initially of  'attraction' instead of 'attractive', but I also think it will be a bit better after my other changes.


DanRoscigno (approved):
Thanks for fixing that!

PR #232, opened 2022-07-19
Introduce the clients earlier for the quickstart guide

yamasite:
I used a headless Ubuntu to run this tutorial, only to find that this tutorial relies on a web UI. The CLI client comes much later in the tutorial, and I have to turn to the web UI to create the database first. So it might be a good idea to introduce the clients earlier and guide developers to run the SQL all the way down?

### DIFF
--- a/docs/en/quick-start.mdx
+++ b/docs/en/quick-start.mdx
@@ -82,6 +82,10 @@ If your OS is not supported or for other install options, view the [installation
 
 ## 2. Connect to ClickHouse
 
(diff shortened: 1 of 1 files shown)


DanRoscigno:
I am in the middle of a rewrite of the quickstart, I will take this into consideration.  Thanks @yamasite !

DanRoscigno:
Thanks @yamasite , I will add my rewrites later on, merging yours now.

PR #248, opened 2022-07-21
Removes the home page from /docs

rfraposa:
Including the little "home" icon in the bread crumbs. So /docs now goes to /docs/en/intro and that Docusaurus-built landing page is gone.


DanRoscigno (approved):
Thanks, this is great!

PR #251, opened 2022-07-21
match website light mode footer and sidebar

DanRoscigno:


rfraposa (approved):
Thanks for figuring this out!

PR #260, opened 2022-07-28
udate nav

DanRoscigno:
This moves a new folder of development docs into reference > development.
cc: @gingerwizard


rfraposa:
Question: Should the new URL be `/en/development/native-protocol`? And in that case, why doesn't Docusaurus pick up the nav automatically from the _category_.yml file?

DanRoscigno:
> Question: Should the new URL be `/en/development/native-protocol`? And in that case, why doesn't Docusaurus pick up the nav automatically from the _category_.yml file?

because it is in the clickhouse-docs repo, and the reference > development stuff is in ClickHouse/ClickHouse

PR #265, opened 2022-08-01
Fix hide on scroll

DanRoscigno:
Needed CSS in addition to setting hideOnScroll


rfraposa (approved):

rfraposa:
Brilliant - thanks for figuring this out!

PR #285, opened 2022-08-10
simplify tutorial

DanRoscigno:


rfraposa:
Thanks Dan. Sad to lose the Dictionary part of this, but for the Cloud it just won't make sense, so this LGTM.

PR #290, opened 2022-08-12
sample method for including reusable content, do not merge

DanRoscigno:
Some content in the docs, for example "you need the hostname, username, password for you ClickHouse service to follow this tutorial.  You can collect this information from XYZ."  

Steps:
- create a *snippet* file that includes the reusable markdown
- place the file in a dir within the `docs` folder that starts with an underscore (directories starting with underscore are ignored by when building the navigation)
- name, or rename the doc that needs to include the content from `.md` to `.mdx` extension


docs/en/guides/sre/sample.mdx
```diff
@@ -0,0 +1,29 @@
...
+## Prerequisites
+
+- step
+- step
+- step
+- step
+
+import ConnectionDetails from '@site/docs/en/_snippets/_gather_your_details_native.md';
+
+<ConnectionDetails/>;
```
DanRoscigno, line 21:
```suggestion
<ConnectionDetails/>
```

PR #299, opened 2022-08-15
Fixing missing logo in firefox.

gjones:
### Summary 
The ClickHouse logo was not displaying in FireFox. Additionally the width of the header was not reaching the full width of the page. This PR fixes both of these issues. 

#### Before
CleanShot 2022-08-15 at 13 48 53@2x

#### After
CleanShot 2022-08-15 at 13 49 02@2x

Sidenote: We were using `-webkit-fill-available` which isn't supported by firefox browsers, I've been investigating why we were using it but could not find a good reason. I've replaced it with `width:100%` and have tested across browsers and device sizes and don't see any visual degradation. 

### Testing
- [x] Light & Dark mode
- [x] Chrome
- [x] Safari
- [x] Firefox 
- [x] Mobile safari (via simulator) 
- [x] Major responsive breakpoint sizes

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -1,7 +1,7 @@
 /* keep the navbar from disappearing when scrolling */
 .navbar--fixed-top {
   position: fixed;
-  width: -webkit-fill-available;
+  width: 100%;
 } 
 
 /* give some sace for the breadcrumbs */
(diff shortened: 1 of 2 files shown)


DanRoscigno:
Thanks @gjones I was using the `-webkit-fill-available` as without it I was seeing the search bar shortened like was reported in Firefox.  Thanks for fixing this up!!

DanRoscigno (approved):
LGTM

PR #303, opened 2022-08-16
Adding exact height, width to logo SVG

gjones:
Related to: https://github.com/ClickHouse/clickhouse-docs/pull/299 which didn't produce the desired result. 

After a little more research I noticed that Firefox needs the height and width to be a little more verbose than other browsers. While the sizing and viewBox provided in the logo SVG itself worked locally, something in the build process meant that in production. the sizes weren't being recognised. This is fixed by verbosely adding desired height and width via css.

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -35,6 +35,12 @@
 .menu__caret:before {
     filter: none !important;
 }
+
+.navbar__logo {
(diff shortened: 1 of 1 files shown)


DanRoscigno (approved):
Thanks again @gjones !

PR #310, opened 2022-08-18
Fixed tables width for integrations

Ryado:
I reused a CSS class that was created for the adopters table to fix the table width
`<div class="adopters-table">`


DanRoscigno:
can you test with the nbsp padding removed?

DanRoscigno:
it breaks the right hand side bar.  Scroll the page and the right hand nav is overlapping the table.  I guess it works in Adopters because there are no H2 headings.

PR #327, opened 2022-08-25
Fix missing CREATE TABLE statement and add sql wrapper

Raffaellorr:
I found that there isn't CREATE TABLE statement in [utilizing-materialized-views](https://clickhouse.com/docs/en/integrations/kafka/kafka-table-engine#2-utilizing-materialized-views), it would cause the Syntax error.


Raffaellorr:
I found that there isn't `CREATE TABLE` statement in [utilizing-materialized-views](https://clickhouse.com/docs/en/integrations/kafka/kafka-table-engine#2-utilizing-materialized-views), it would cause the `Syntax error`.

PR #345, opened 2022-08-31
Fix link is invalid.

FannyFirst:
Changed some link like "..integrations/kafka/.." to "..integrations/data-ingestion/kafka/.."


DanRoscigno:
Thanks for catching this @FannyFirst .  I pushed some updates to switch to relative links.

PR #360, opened 2022-09-07
Add note about delete from

joshuataylor:
ClickHouse 22.8 supports `DELETE FROM`, which should be noted for users.


DanRoscigno:
@rfraposa can you compare this with your PR and combine?

joshuataylor:
I'm so sorry, I missed this message.

PR #380, opened 2022-09-15
update s3 example using NYC Taxi data

DanRoscigno:
- fix bucket URL
- fix list at end of doc
- change query, the records in this subset only have unknown payment type, so the old query is not interesting.


rfraposa (approved):

rfraposa:
Does the schema work too (is the date issue fixed)?

DanRoscigno:
> Does the schema work too (is the date issue fixed)?

This doc does not use inference, the schema is defined in the query.  Only when using inference is there an issue.

PR #382, opened 2022-09-18
Update performance.md

tbragin:
Instead of referring to the "internal Yandex results", point to https://benchmark.clickhouse.com/ and ClickBench.

cc @DanRoscigno


DanRoscigno (approved):
LGTM, thanks for updating this!

PR #410, opened 2022-10-02
add nodejs client docs

mshustov:
The PR is based on the existing https://github.com/ClickHouse/clickhouse-js#readme with some minor changes.
cc @slvrtrn


DanRoscigno:
@mshustov is the client also compatible with node 18?  If so I will drop the mention of 14 and 16 and just say `The client is compatible with all the [maintained](https://github.com/nodejs/release#readme) nodejs LTS versions` since 18 goes into maintenance in 3 weeks time.

mshustov:
@DanRoscigno yes, we test against `v18`. I didn't add it because it's not in the official maintenance yet.

PR #416, opened 2022-10-04
update go client docs

gingerwizard:
@DanRoscigno this depends on https://github.com/ClickHouse/clickhouse-go/pull/776 and release of client. Please don't merge till I 👍


docs/en/integrations/language-clients/go/clickhouse-go/performance-tips.md
```diff
@@ -9,6 +9,7 @@ description: Performance Tips
 # Performance Tips
 
 * Utilize the ClickHouse API where possible, especially for primitive types. This avoids significant reflection and indirection.
+* If reading large datasets considering modifying the [BlockBufferSize](clickhouse-api#connection-settings). This will increase the memory footprint but will mean more blocks can decoded in parallel during row iteration. The default value of 2 is conservative and minimizes memory overhead. Higher values will mean more blocks in memory. This requires testing since different queries can produce different block sizes. It can therefore be set on a [query level](clickhouse-api#using-context) via the Context.
```
DanRoscigno, line 12:
```suggestion
* If reading large datasets consider modifying the [BlockBufferSize](clickhouse-api#connection-settings). This will increase the memory footprint but will mean more blocks can be decoded in parallel during row iteration. The default value of 2 is conservative and minimizes memory overhead. Higher values will mean more blocks in memory. This requires testing since different queries can produce different block sizes. It can therefore be set on a [query level](clickhouse-api#using-context) via the Context.
```

mshustov (approved):

PR #419, opened 2022-10-04
init datagrip docs

DanRoscigno:


DanRoscigno:
from @vadimp 

For now what is important is that the user will see the Host field should contain https://... string.

May be it's worth adding a reference to clickhouse-jdbc project for more information about JDBC string configuration, to cover all changes in the future?
For more JDBC URL setting, please refer to https://github.com/ClickHouse/clickhouse-jdbc

DanRoscigno:
cc: @Ryado added to integrations page

DanRoscigno:
image

vadimp:
There is a validation error in the log, saying:
```
[ERROR] @docusaurus/plugin-client-redirects: some redirects would override existing paths, and will be ignored: 
- {"from":"/en/getting-started/install","to":"/en/install"}
```
Besides that, I'm happy with the current doc. Now it's straightforward.

DanRoscigno:
That error is harmless, I have a PR to fix it in review.  Merging.  Thanks very much for your help!

PR #440, opened 2022-10-05
Update Node.js docs for 0.0.5 release

slvrtrn:
* Updated API
* Added a bit more examples and know-how
* Reformat the document
* Fix code snippets so those are valid TypeScript and more copy-paste friendly
* Spelling and wording fixes

### DIFF
--- a/docs/en/integrations/language-clients/nodejs.md
+++ b/docs/en/integrations/language-clients/nodejs.md
@@ -7,82 +7,113 @@ description: The official Node.js client for connecting to ClickHouse.
 ---
 
 # ClickHouse JS
+
 The official Node.js client for connecting to ClickHouse.
 The client is written in TypeScript and provides typings for the client public API.
 
 ## Environment requirements
-The nodejs must be available in the environment to run the client.
-The client is compatible with all the [maintained](https://github.com/nodejs/release#readme) nodejs LTS versions: `v14.x` and `v16.x`.
 
-As soon as a nodejs version approaches End-Of-Life, the client drops support for the outdated insecure nodejs version.
+The Node.js must be available in the environment to run the client.
+The client is compatible with all the [maintained](https://github.com/nodejs/release#readme) Node.js LTS
+versions: `v14.x` and `v16.x`, as well as the current `v18.x`.
+
+As soon as a Node.js version approaches End-Of-Life, the client drops support for it as it is considered outdated and
+insecure.
 
 *Note*: The Browser environment is not officially supported at the moment.
 
 ## Installation
+
 To install the latest available client version, run:
+
 ```sh
 npm i @clickhouse/client
 ```
+
 ## Support policy
-Clients provide forward compatibility with all the ClickHouse versions released in the following **2 years** from the client release date.
-The clients also provide backward compatibility with all the ClickHouse versions supported when a client was released, which maps to 3 latest `stable` and max 2 `lts` releases.
+
+Clients provide forward compatibility with all the ClickHouse versions released in the following **2 years** from the
+client release date.
+The clients also provide backward compatibility with all the ClickHouse versions supported when a client was released,
+which maps to 3 latest `stable` and max 2 `lts` releases.
 
 ## Compatibility with ClickHouse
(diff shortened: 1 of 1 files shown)


docs/en/integrations/language-clients/nodejs.md
```diff
@@ -7,82 +7,113 @@ description: The official Node.js client for connecting to ClickHouse.
...
+
 The official Node.js client for connecting to ClickHouse.
 The client is written in TypeScript and provides typings for the client public API.
 
 ## Environment requirements
-The nodejs must be available in the environment to run the client.
-The client is compatible with all the [maintained](https://github.com/nodejs/release#readme) nodejs LTS versions: `v14.x` and `v16.x`.
 
-As soon as a nodejs version approaches End-Of-Life, the client drops support for the outdated insecure nodejs version.
+The Node.js must be available in the environment to run the client.
```
DanRoscigno, line 16:
```suggestion
Node.js must be available in the environment to run the client.
```

docs/en/integrations/language-clients/nodejs.md
```diff
@@ -7,82 +7,113 @@ description: The official Node.js client for connecting to ClickHouse.
...
+The client is compatible with all the [maintained](https://github.com/nodejs/release#readme) Node.js LTS
+versions: `v14.x` and `v16.x`, as well as the current `v18.x`.
```
DanRoscigno, line 18:
```suggestion
versions.
```

DanRoscigno: up to you, but you could just say LTS versions and not have to worry about updating the doc when 14 drop s off.

docs/en/integrations/language-clients/nodejs.md
```diff
@@ -7,82 +7,113 @@ description: The official Node.js client for connecting to ClickHouse.
...
+
 ```sh
 npm i @clickhouse/client
 ```
+
 ## Support policy
-Clients provide forward compatibility with all the ClickHouse versions released in the following **2 years** from the client release date.
-The clients also provide backward compatibility with all the ClickHouse versions supported when a client was released, which maps to 3 latest `stable` and max 2 `lts` releases.
+
+Clients provide forward compatibility with all the ClickHouse versions released in the following **2 years** from the
```
DanRoscigno, line 35: I think 2 years is a long time, is this confirmed with the support team?  And the wording for backward means, to me, that there is support for unsupported versions of ClickHouse.  In 2024 the current version of the client will support versions all the way back to 2021, but the clickhouse policy will only be for the lts version from 2023 and 2024.

DanRoscigno (commented):
made some comments, I will finish in a bit.

docs/en/integrations/language-clients/nodejs.md
slvrtrn: I think we should just remove it indeed.

docs/en/integrations/language-clients/nodejs.md
slvrtrn: cc @mshustov

DanRoscigno (approved):
LGTM, I tested connecting to and creating a table against Cloud.  Please talk with your team and maybe @melvynator about the support statement, I don't know if the policy is different for clients than it is for server.

docs/en/integrations/language-clients/nodejs.md
mshustov: > even though some of the old versions of ClickHouse server that were supported at the time of the client release are no longer supported.

Let's add the following statement from the proposal: `Once a ClickHouse version is no longer supported, support for this version in language clients is also dropped.` @DanRoscigno WDYT?

DanRoscigno: That works!

slvrtrn: I removed the entire section for the time being.

DanRoscigno: good thinking, once the policy gets published we can just link.

PR #448, opened 2022-10-06
Warning that Automatic deduplication is disabled for asynchronous inserts

tom-clickhouse:
Warning that Automatic deduplication is disabled for asynchronous inserts

### DIFF
--- a/docs/en/manage/tuning-for-cloud-cost-efficiency.md
+++ b/docs/en/manage/tuning-for-cloud-cost-efficiency.md
@@ -58,6 +58,15 @@ As an example, this is how you can do that within a JDBC connection string when 
 ```
 
 
+:::note Automatic deduplication is disabled when using asynchronous inserts
+Manual batching (see [section above](#ingest-data-in-bulk))) has the advantage that it supports the [built-in automatic deduplication](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication/) 
+of table data if (exactly) the same insert statement is send mutliple times to ClickHouse Cloud, 
+for example because of an automatic retry in a client software because of some temporarily network connection issues.
+
+Asynchronous inserts don't support this built-in automatic deduplication of table data in case of insert statement retries.
+:::
+
+
 ## Use a low cardinality partitioning key
 
 When you send an insert statement (that should contain many rows - see [section above](#ingest-data-in-bulk)) to a table in ClickHouse Cloud, and that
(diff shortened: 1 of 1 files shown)


docs/en/manage/tuning-for-cloud-cost-efficiency.md
```diff
@@ -58,6 +58,15 @@ As an example, this is how you can do that within a JDBC connection string when
 ```
 
 
+:::note Automatic deduplication is disabled when using asynchronous inserts
+Manual batching (see [section above](#ingest-data-in-bulk))) has the advantage that it supports the [built-in automatic deduplication](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication/) 
+of table data if (exactly) the same insert statement is send mutliple times to ClickHouse Cloud,
```
DanRoscigno, line 63:
```suggestion
of table data if (exactly) the same insert statement is sent multiple times to ClickHouse Cloud, 
```

docs/en/manage/tuning-for-cloud-cost-efficiency.md
```diff
@@ -58,6 +58,15 @@ As an example, this is how you can do that within a JDBC connection string when
 ```
 
 
+:::note Automatic deduplication is disabled when using asynchronous inserts
+Manual batching (see [section above](#ingest-data-in-bulk))) has the advantage that it supports the [built-in automatic deduplication](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication/) 
+of table data if (exactly) the same insert statement is send mutliple times to ClickHouse Cloud, 
+for example because of an automatic retry in a client software because of some temporarily network connection issues.
```
DanRoscigno, line 64:
```suggestion
for example because of an automatic retry in client software because of some temporary network connection issues.
```

docs/en/manage/tuning-for-cloud-cost-efficiency.md
```diff
@@ -58,6 +58,15 @@ As an example, this is how you can do that within a JDBC connection string when
...
+
+Asynchronous inserts don't support this built-in automatic deduplication of table data in case of insert statement retries.
```
DanRoscigno, line 66: I think maybe drop the `in case of insert statement retries`.  I think dedup is not done no matter the reason if using async.

DanRoscigno (changes requested):
some changes and a question

docs/en/manage/tuning-for-cloud-cost-efficiency.md
tom-clickhouse: Makes sense, let's drop.

DanRoscigno (approved):

PR #467, opened 2022-10-17
Add ports to AWS PrivateLink

DanRoscigno:


melvynator:
I would also add which IP address to add in the IP filtering section for more clarity. WDYT?

DanRoscigno:
> I would also add which IP address to add in the IP filtering section for more clarity. WDYT?

I think it is already highlighted with a note and red rectangle, isn't it? If not, can you add it?

PR #475, opened 2022-10-19
add try for free link

DanRoscigno:
image


okayfrosty:
@DanRoscigno I'm good with this. I actually like the arrow box, hopefully that makes community users feel less like we're baiting and switching them since it's commonly used to indicate "you're going somewhere else now." Let's see how it goes!

DanRoscigno:
right on, thanks.  Initially, I had a lowercase `free`, it is in the process of rebuilding with uppercase `Free`.  Have a good one!

PR #488, opened 2022-10-25
October 25th release notes

DanRoscigno:
image

### DIFF
--- a/docs/en/whats-new/changelog/cloud.md
+++ b/docs/en/whats-new/changelog/cloud.md
@@ -5,9 +5,13 @@ sidebar_label:  Cloud
 title: Cloud Changelog
 ---
 
-## October 25, 2022
+## October 4, 2022
+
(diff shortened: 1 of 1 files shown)


tbragin (commented):
@DanRoscigno I think we should leave the first entry in the change log as "October 4, 2022" and maybe have some statement about the fact that we launched our offering then, and point to release blog? What do you think?

DanRoscigno:
> @DanRoscigno I think we should leave the first entry in the change log as "October 4, 2022" and maybe have some statement about the fact that we launched our offering then, and point to release blog? What do you think?

@tbragin Typically, a changelog has the most recent info at the very top.  I would like to keep the October 4th entry pinned at the top for a while, and then have the rest of the content flow below with the most recent release at the top.  Do you think this is OK?  Maybe we will change the header for the Oct 4th entry to be something other than the date to differentiate it from the rest?

image

tbragin (commented):
@DanRoscigno LGTM, except let's reverse ordering, with newer changes at the top (so Oct 25 at the top, Oct 4 at the bottom), like we do with other changelogs.

PR #502, opened 2022-10-31
Favicons don't work anyway

alexey-milovidov:
Favicons were broken while migrating to the new docs.
I don't see anyone cares.

This PRs removing the garbage.


blinkov:
This was the only reason why those links were in HTML instead of Markdown - so it can be changed back for consistency too.

DanRoscigno:
What did that class do?  Pull the icon from the 3rd party website?

alexey-milovidov:
> Pull the icon from the 3rd party website?

Yes.

PR #503, opened 2022-10-31
add compatibility setting

DanRoscigno:


nikitamikhaylov:
The compatibility setting will just set some other specified settings to a value which was in a particular ClickHouse release. An example of them is listed in our code itself. https://github.com/ClickHouse/ClickHouse/blob/9e1627fcdd27a3bf6719598aa9c30d4bccf304be/src/Core/SettingsChangesHistory.h#L79-L114

DanRoscigno:
I will test, it did succeed in Cloud, but I did not know how to check it.  I will look in system.settings, and then stop and start the cluster.  This is the way we recommend that people set `async_insert` in Cloud, so I think it does work.

DanRoscigno:
I set this a week ago, and it has survived pauses and restarts:
image

Today I used Action > Stop and then Action > Start to restart the service and the change has also survived.  I will add the SQL to verify the change, and then merge, OK @nikitamikhaylov ?

nikitamikhaylov:
Cool! The reason I asked - I didn't understand where we store that knowledge, but seems like it is stored in Keeper and everything Ok. Let's merge!

nikitamikhaylov (approved):

PR #504, opened 2022-10-31
remove early access guide, link to compatibility

DanRoscigno:
Removes Early Access guide, replaces with a redirect to What's New > Cloud Compatibility


rfraposa (approved):
Thanks for removing this!

PR #506, opened 2022-11-01
Adding details for CVE-2022-44011 and CVE-2022-44010

santrancisco:
Previously there were 2 heap buffer overflow vulnerabilities that was reported to us through our Bugbounty program. 

Core team immediately jumped on it within 2 days and addressed the issue. 

It took a while to get the CVE submission through but we have them now for tracking and to credit the original researcher.


rfraposa (approved):
LGTM

PR #527, opened 2022-11-03
Update cloud.md

tbragin:
@DanRoscigno after looking at the formatting, I think this is slightly better, and still makes sense. Let me know if you agree.


DanRoscigno (approved):
LGTM

DanRoscigno:
Thanks @tbragin

PR #528, opened 2022-11-04
Update tutorial

DanRoscigno:
Add dictionaries back in.  I will hold this until we are ready, but if you could review it early I would appreciate it.  Thanks!

### DIFF
--- a/docs/en/tutorial.md
+++ b/docs/en/tutorial.md
@@ -2,7 +2,7 @@
 slug: /en/tutorial
 sidebar_label: ClickHouse Tutorial
 sidebar_position: 0.5
-keywords: [clickhouse, install, tutorial]
+keywords: [clickhouse, install, tutorial, dictionary, dictionaries]
 ---
 import SQLConsoleDetail from '@site/docs/en/_snippets/_launch_sql_console.md';
 
@@ -10,7 +10,7 @@ import SQLConsoleDetail from '@site/docs/en/_snippets/_launch_sql_console.md';
 
 ## What to Expect from This Tutorial?
 
-In this tutorial, you will create a table and insert a large dataset (two million rows of the [New York taxi data](./getting-started/example-datasets/nyc-taxi.md)). Then you will run queries on the dataset.
+In this tutorial, you will create a table and insert a large dataset (two million rows of the [New York taxi data](./getting-started/example-datasets/nyc-taxi.md)). Then you will run queries on the dataset, including an example of how to create a dictionary and use it to perform a JOIN.
 
 :::note
 This tutorial assumes you have access to a running ClickHouse service.  If not, check out the [Quick Start](./quick-start.mdx).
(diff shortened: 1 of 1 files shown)


docs/en/tutorial.md
```diff
@@ -341,7 +341,169 @@ Let's run some queries to analyze the 2M rows of data...
     │ 2015-07-01 00:41:48 │ 2015-07-01 00:44:45 │          6.3 │                 -94 │                  132 │ JFK          │ 2015 │   1 │    0 │
     │ 2015-07-01 01:06:18 │ 2015-07-01 01:14:43 │        11.76 │                  37 │                  132 │ JFK          │ 2015 │   1 │    1 │
     ```
+## 4. Create a Dictionary
 
+If you are new to ClickHouse, it is important to understand how ***dictionaries*** work. A dictionary is a mapping of key->value pairs that is stored in memory.
```
nikitamikhaylov, line 346: Maybe somehow to cover that a key could be both simple (a UInt64) and complex (multiple values). A value is actually multiple attributes. But in general Ok

docs/en/tutorial.md
```diff
@@ -341,7 +341,169 @@ Let's run some queries to analyze the 2M rows of data...
...
+  CREATE DICTIONARY taxi_zone_dictionary
+  (
+    LocationID UInt16,
+    Borough String,
+    Zone String,
+    service_zone String
+  )
+  PRIMARY KEY LocationID
+  SOURCE(CLICKHOUSE(TABLE 'taxi_zone_source'))
+  LAYOUT(HASHED_ARRAY())
```
nikitamikhaylov, line 389: Some link to another possible layouts?

docs/en/tutorial.md
```diff
@@ -341,7 +341,169 @@ Let's run some queries to analyze the 2M rows of data...
...
+  ```sql
+  CREATE DICTIONARY taxi_zone_dictionary
+  (
+    LocationID UInt16,
+    Borough String,
+    Zone String,
+    service_zone String
+  )
+  PRIMARY KEY LocationID
+  SOURCE(CLICKHOUSE(TABLE 'taxi_zone_source'))
```
nikitamikhaylov, line 388: Maybe also to cover that only user `default` can create a dictionary without specifying a password? If a `default` user will create another user then it is required to specify it like this `USER another_user PASSWORD plain_text`

docs/en/tutorial.md
```diff
@@ -341,7 +341,169 @@ Let's run some queries to analyze the 2M rows of data...
...
+
+Let's write some queries that join the `taxi_zone_dictionary` with your `trips` table.
+
+1. We can start with a simple JOIN that acts similarly to the previous airport query above:
+    ```sql
+    SELECT
+        count(1) AS total,
+        Borough
+    FROM trips
+    JOIN taxi_zone_dictionary ON toUInt64(trips.pickup_nyct2010_gid) = taxi_zone_dictionary.LocationID
```
nikitamikhaylov, line 472: Maybe add another queries to compare the speed? For customer to understand why he might need dictionaries? https://pastila.nl/?0007b917/37e107b6bf13dcd9507cb89e2da6faf4  There are might be some type conversion problems, but for quick demo it is Ok I think

nikitamikhaylov (approved):

DanRoscigno:
@nikitamikhaylov I will add additional queries to the tutorial when I come back from vacation, I opened https://github.com/ClickHouse/clickhouse-docs/issues/552

Thanks for all of the suggestions!

PR #534, opened 2022-11-05
Update _supported_regions.md

rfraposa:
I'm not a fan of the "performance will vary" comment because it doesn't explain why. I would suggest simply removing it, but open to other suggestions. If we keep it, let's just make it an asterisk (and remove the hyperlinking)


DanRoscigno:
@krithika-ch please review

krithika-ch:
Sorry missed this. LGTM

PR #535, opened 2022-11-06
Update production.md

rfraposa:
I just discovered this page - it's a good read. I cleaned up a few typos.


docs/en/faq/operations/production.md
```diff
@@ -9,32 +9,32 @@ toc_priority: 10
 
 First of all, let’s discuss why people ask this question in the first place. There are two key reasons:
 
-1.  ClickHouse is developed with pretty high velocity and usually, there are 10+ stable releases per year. It makes a wide range of releases to choose from, which is not so trivial choice.
+1.  ClickHouse is developed with pretty high velocity and usually, there are 10+ stable releases per year. That makes a wide range of releases to choose from, which is not so trivial of a choice.
```
DanRoscigno, line 12:
```suggestion
1.  ClickHouse is developed with pretty high velocity, and usually there are 10+ stable releases per year. That makes a wide range of releases to choose from, which is not so trivial of a choice.
```

PR #538, opened 2022-11-08
add cloud specifics for admin role / user

DanRoscigno:
Instructions for creating an admin user.  The workflow might be something like this:
- Create a ClickHouse Cloud service
- Create a non-default admin role
- Create a database for a project
- Create a user for a project, and assign the admin role for the project DB to the project user
- Invite a Developer user to the org

### DIFF
--- a/docs/en/guides/sre/user-management/users-and-roles.md
+++ b/docs/en/guides/sre/user-management/users-and-roles.md
@@ -2,15 +2,150 @@
 sidebar_label: Defining SQL Users and Roles
 sidebar_position: 20
 slug: /en/guides/sre/users-and-roles
+hide_table_of_contents: true
 ---
+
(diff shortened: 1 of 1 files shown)


DanRoscigno:
Preview:
image

e-mars:
@DanRoscigno looks good. i would recommend to add a note that if a new database is added, must log in with default user and run the grants for the admin role to the new database. may avoid some confusion with users since they might believe that the initial grants will apply to the new databases.

e-mars (approved):
added comment.

DanRoscigno:
> @DanRoscigno looks good. i would recommend to add a note that if a new database is added, must log in with default user and run the grants for the admin role to the new database. may avoid some confusion with users since they might believe that the initial grants will apply to the new databases.

Thanks @e-mars added:

image

PR #556, opened 2022-11-16
Update s3-table-engine with a non-AWS example

antoniojtorres:
I was recently helped in the issue queue with regards to mounting a wasabi bucket as a disk, I was having difficulty troubleshooting the error I was getting when mapping the AWS style URLs to the Wasabi style (https://github.com/ClickHouse/ClickHouse/issues/42043#issuecomment-1309638255). The fix I needed seems worthy of adding to the docs. 

I added a small example for wasabi (and likely other users) that explains that both the bucket and the folder need to be added to the end of the endpoint path.

I think it would be useful to compile additional examples for other S3 compatible storage systems. 

Thank you for the wonderful docs!


DanRoscigno:
@melvynator @alexey-milovidov are you ok with a wasabi example in the docs?  I am guessing that we do not test against wasabi as I cannot find it in the `/tests` folder.

DanRoscigno (approved):
Thanks @antoniojtorres !

PR #557, opened 2022-11-16
Remove duplicate create statement in configuring-s3-for-clickhouse-use.md

antoniojtorres:
The configuring s3 for clickhouse doc seems to have two separate create statements. Looks to be a duplicate. Removed the extra create statement.


DanRoscigno (approved):
Thanks Antonio!

PR #575, opened 2022-11-28
Add create table instructions to Redshift to ClickHouse migration

cwurm:
Adds an intermediary step to https://clickhouse.com/docs/en/integrations/redshift/redshift-pivot-to-clickhouse/, creating the correct table structure before inserting data.


DanRoscigno:
Thanks for doing this @cwurm !

PR #593, opened 2022-12-01
move Redshift examples out of ETL

DanRoscigno:
The Reshift examples are three different techniques for migrating data from Redshift to ClickHouse.  Placing this under ETL tools was confusing, as the migration techniques don't actually involve ETC (well, one kind of does, but the other two do not).

This PR moves the Redshift examples out of the ETL tools category:

image

@rfraposa please take note of this when you re-record the migration video.


rfraposa:
Actually, I would remove the word "Example:" and just have "Migrate from Redshift", but it's not a deal-breaker

DanRoscigno:
> Actually, I would remove the word "Example:" and just have "Migrate from Redshift", but it's not a deal-breaker

I think the pages are good examples of the techniques, in addition to the concrete example of moving from Redshift.  I want people who are migrating from services other than Redshift to read them also.

PR #606, opened 2022-12-07
Update _supported_regions.md

tbragin:
@DanRoscigno Updating this snippet because we now enabled ap-south-1 for self-service for Production services. On the other hand ap-southeast-1 and ap-south-1 currently do not support Development services.

cc @krithika-ch FYI


DanRoscigno:
Thanks @tbragin !

PR #609, opened 2022-12-07
use new default_role for SQL user doc

DanRoscigno:
@e-mars ready for testing.  I tested the Cloud flow.  Thanks so much for this great information.

image


e-mars (commented):
lgtm

PR #612, opened 2022-12-09
add ssl info to replication

### LINKED ISSUE #608: Link SSL-TLS doc under Replication and Sharding section
Since the doc `Configuring SSL-TLS` is very similar to the `Configuring ClickHouse Keeper` but with a different example of how to configure and test replication, I think it also belongs in this section, `Replication and Sharding` also.

https://clickhouse.com/docs/en/guides/sre/configuring-ssl

DanRoscigno:
closes #608


e-mars (approved):
thanks! lgtm

PR #614, opened 2022-12-09
basic jupyter use

DanRoscigno:


docs/en/integrations/language-clients/python/jupyter.md
```diff
@@ -0,0 +1,78 @@
...
+import clickhouse_connect
+client = clickhouse_connect.get_client(
+    host=ch_host,
+    port=ch_port,
+    secure='true',
+    username=ch_username,
+    password=ch_password)
+```
+```python
+result = client.query('SELECT query FROM system.query_log ORDER BY event_date DESC LIMIT 1')
```
genzgd, line 73: I think you can use `client.query_df` directly which should be more efficient?

DanRoscigno: Can you give me an example?

genzgd: `client.query_df` should work exactly the same, but it returns a Pandas Dataframe that I *think* Juptyer can just use directly with any more code.  `client.query` returns a "QueryResult" object and you need some manipulation to use it directly in the Notebook.

PR #615, opened 2022-12-13
Add aws marketplace

DanRoscigno:
@kapustor I changed "AWS Marketplace" to "aws marketplace" everywhere as that is the case I see AWS using on the marketplace site.  Let me know if that is OK.

image

Here is where it is in the menu:
image

and here:

image


kapustor:
Hi @DanRoscigno , looks good, but just a few things:

- can we please rename the page from "AWS Marketplace" to "Marketplace" as later we will add here GCP and Azure information
- I see a section "What are the best practices?" in the end of the page, is it supposed to be here? Seems like it is not connected to marketplace

kapustor (commented):
Commented above

PR #624, opened 2022-12-15
Replace external kafka option by Buffer table engine

CamiloSierraH:
The buffer table engine is a great option when you are sending a large number of insert request with a low number of rows on each.


DanRoscigno:
++

PR #629, opened 2022-12-16
Mirror best practices in Optimize

DanRoscigno:
Adds the best practices from Cloud to the Optimize section, as the Cloud techniques are also applicable for self-managed:
image


rfraposa:
Looks good, but let's discuss on Monday when we meet so I have more time to review. Thanks!

PR #634, opened 2022-12-20
add public list of egress IP address for CLickHouse Cloud

CamiloSierraH:
I have removed in MySQL and PostgreSQL the message that says it was not ready on ClickHouse cloud as this has been released, and I could test it.
I have also added the list of public egress IPs


mshustov (approved):

docs/en/cloud/security/ip-egress-traffic-list.md
```diff
@@ -0,0 +1,50 @@
...
+slug: /en/manage/security/ip-egress-traffic-list
+sidebar_label: IP Egress Traffic List
+title: IP Egress Traffic List
+---
+
+## Integrations could require egress access
+
+If you are using integration like the MySQL or PostgreSQL Engine, It's possible that you need to authorize ClickHouse Cloud to access to your instances.
+
+You could use this list of public IPs in `firewalls` or `Authorized networks` in GCP or in `Security Groups` in AWS or in any other infrastrucutre egrees management system you are using.
```
mshustov, line 11: Can we add screenshots to help users to find these?

mshustov, line 11: Should we add a comment for Azure?

CamiloSierraH: I have no experience on Azure, let me see what can I find on the net !

CamiloSierraH: I have this one for GCP, do we want one for AWS too ?
Screenshot 2022-12-20 at 12 25 03

DanRoscigno: Should there be a note about not allowing these public IPs if using AWS  PrivateLink, or would the person managing this know to ignore this if they are using PrivateLink? (I guess they would know that, right?)

CamiloSierraH: Hey @DanRoscigno , that is a good catch, I think we should add something about it, we could just do a small `note` about it?

DanRoscigno (approved):
Thanks, this is great!

PR #637, opened 2022-12-23
Update clickhouse-client-local.md

alexey-milovidov:

PR #638, opened 2022-12-23
move clickhouse-client and install up

DanRoscigno:
@guykoh like so?
image


guykoh (approved):
Looks much better IMO.
Thanks for doing this @DanRoscigno

DanRoscigno:
closing in favor of #637 

@guykoh Alexey had started a PR just before this one, I am combining them.  Note that his changes the process from installing the client to using the curl | sh method, so `clickhouse-client` becomes `clickhouse client`.  I will make a note in the discussion with design.

PR #646, opened 2022-12-30
Missed <remote_servers> tag

yakov-olkhovskiy:


DanRoscigno:
Thanks for this @yakov-olkhovskiy , I just looked at the configs I use when testing against replicated systems and I include the `<clickhouse> and also specify replace=true for remote servers (to remove the example ones in the default config).  Do you think we should present these config files like this?  I am going to be reviewing the configuration section of the docs this coming week and will make these changes there also if you agree.

```
<clickhouse>
    <remote_servers replace="true">
        <cluster_2S_1R>
...
        </cluster_2S_1R>
    </remote_servers>
</clickhouse>
```

yakov-olkhovskiy:
I think in this particular place it's assumed that this block should be put inside `<clickhouse>` - from all the previous instructions, and `replace="true"` is not necessary because no other overlapping configs are assumed in the example

DanRoscigno (approved):
LGTM

PR #652, opened 2023-01-04
Fix trip duration query in the first Clickhouse tutorial

AntonRutkevich:
Current query seems to be grouping by hour, not by minute. 

Indeed, out of 2 million rows in the table, only ~14K have this value above 0. 
```
SELECT count()
FROM trips
WHERE truncate(dateDiff('second', pickup_datetime, dropoff_datetime) / 3600) > 0

Query id: e1c973b4-cda6-4917-86f3-45501a1b4049

┌─count()─┐
│   13932 │
└─────────┘

1 row in set. Elapsed: 0.023 sec. Processed 2.00 million rows, 16.00 MB (88.67 million rows/s., 709.36 MB/s.)
```

### DIFF
--- a/docs/en/tutorial.md
+++ b/docs/en/tutorial.md
@@ -237,7 +237,7 @@ Let's run some queries to analyze the 2M rows of data...
         avg(fare_amount) AS avg_fare,
         avg(passenger_count) AS avg_passenger,
         count() AS count,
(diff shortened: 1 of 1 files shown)


DanRoscigno:
Thanks for catching that Anton!

PR #654, opened 2023-01-05
Fix docs links

victoryNap:
Looks like some of the redirects from `https://clickhouse.com/docs/{section}` don't work properly, while others do. A better fix would be to redirect and use the user, like `https://clickhouse.com/docs/quick-start` 

Updated the following links:
`https://clickhouse.com/docs/tutorial` -> `https://clickhouse.com/docs/en/tutorial` `https://clickhouse.com/docs/connect-a-ui` -> `https://clickhouse.com/docs/en/connect-a-ui` `https://clickhouse.com/docs/guides` -> `https://clickhouse.com/docs/en/guides`

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,10 +27,10 @@ ClickHouse is blazing fast, but understanding ClickHouse and using it effectivel
 
 Some of the highlights of the ClickHouse docs include:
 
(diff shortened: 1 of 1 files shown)


DanRoscigno:
Thanks @victoryNap !

PR #655, opened 2023-01-05
Update JSON doc

DanRoscigno:
I need to run through this again, and maybe it needs to wait for your recent schema inference PR to get released?


Avogar:
I would change it in a bit different way. Let's keep it as it was, but add `SETTINGS allow_experimental_object_type=1` and after that add same queries with `SETTINGS input_format_json_read_objects_as_strings=1` to show that we can load such dataset without using JSON object type.

Avogar:
BTW, enabling `input_format_json_read_objects_as_strings` by default was already merged, but not released to cloud.
Also after https://github.com/ClickHouse/ClickHouse/pull/44501 will be released, format `JSONAsObject` (mentioned in this doc) won't work without enabled `allow_experimental_object_type` (previousely using this format users could use JSON object type without enabling setting, it was a bug)

DanRoscigno:
Adds a page "Loading JSON in 5 steps"

@Avogar can you review the new page json-intro.md please and make sure that I did not use any experimental features, and that I used the right terminology around `nested`.  Thanks!

docs/en/guides/developer/working-with-json/json-load-data.md
```diff
@@ -0,0 +1,129 @@
...
+INSERT INTO http (`@timestamp`, clientip, request.method, request.path, request.version, status, size) SELECT
+    `@timestamp`,
+    clientip,
+    [request['method']],
+    [request['path']],
+    [request['version']],
+    status,
+    size
+FROM s3('https://datasets-documentation.s3.eu-west-3.amazonaws.com/http/documents-01.ndjson.gz', 'JSONEachRow')
+LIMIT 1
```
Avogar, line 63: That's not a good way to use type `Nested`. `Nested` should be used when we have object with nested **arrays** of data. For example:
`{"id" : 1, "nested" : {"field1" : [1, 2], "field2" : ["value1", "value2"]}}`.
It doesn't make sense to use when we have just simple nested object with string values as we have here.
If we are sure that `request` field always have only these three nested fields `method` and `path`, `version` and they are always String, we can:
1) Just leave it as a Map type:
https://pastila.nl/?0007b917/57c1b3d174e4b1f9247777d433eacbd9
2) Create a separate field for each nested fields `method` and `path`, `version`:

 use just a `Tuple(method LowCardinality(String), path String, version LowCardinality(String))`. Like this:

Avogar, line 63: That's not a good way to use type `Nested`. `Nested` should be used when we have object with nested **arrays** of data. For example:
`{"id" : 1, "nested" : {"field1" : [1, 2], "field2" : ["value1", "value2"]}}`.
It doesn't make sense to use when we have just simple nested object with string values as we have here.
If we are sure that `request` field always have only these three nested fields `method` and `path`, `version` and they are always String, we can:
1) Just leave it as a Map type:
https://pastila.nl/?0007b917/57c1b3d174e4b1f9247777d433eacbd9
2) Create a separate field for each nested fields `method` and `path`, `version`:
https://pastila.nl/?0007b917/04ba7ad0d45b04d1513b6d3fa213cc23
3) Use just `Tuple`:
https://pastila.nl/?0007b917/319ff7449765093dbcd97bc9c182bb03

docs/en/guides/developer/working-with-json/json-load-data.md
```diff
@@ -0,0 +1,129 @@
...
+
+This guide walks through the process to load logging data that is 
+in a JSON formatted file in S3.  In order to do this:
+- Examine the file format by selecting one row using the S3 function
+- Create a table to store the data in ClickHouse
+- Load a single row of nested JSON
+- Verify the correct storage of the nested JSON
+- Import the dataset from S3
+
+## Examine the structure of the JSON file
```
Avogar, line 18: Let's maybe also add `DESCRIBE` query? And also worth mentioning that we have not only JSONEachRow format, but also some other JSON formats that can fit users data

Avogar:
> * ephemeral columns can not have data stored in them

Yes

> * ephemeral columns cannot be selected

Yes

> * ephemeral columns are used by other columns as DEFAULT values.  In this case, the value is converted during INSERT to a DateTime for the `timestamp` column

AFAIK yes
> * the `@` symbol is the name from the JSON, it is not anything special for ephemeral columns

Yes, in the JSON data we have field with name `@timestamp`

docs/en/guides/developer/working-with-json/json-semi-structured.md
```diff
@@ -22,30 +22,34 @@ It is important to note that the JSON type primarily syntactically enhances JSON
...
 | name | type | 
 | :--- | :--- | 
 | type | Nullable\(String\) |
-| actor | Object\('json'\) |
-| repo | Object\('json'\) |
+| actor | Nullable(String) |
+| repo | Nullable(String) |
 | created\_at | Nullable\(String\) |
-| payload | Object\('json'\) |
+| payload | Nullable(String) |
```
Avogar, line 37: I still think we should leave it as it was before, but add setting `allow_experimental_object_type=1` and then add same queires with `input_format_json_read_objects_as_strings` to show the user that this dataset can be loaded without using experimental JSON type.

docs/en/guides/developer/working-with-json/json-load-data.md
DanRoscigno: yes, DESCRIBE is needed.

For the other formats, I will add a section in Notes at the bottom.  For beginner tutorials like this, I want to give them a step-by-step guide that just works without a lot of discussion about other options or links to other information until the end, and then at the end, provide more information.  I will look at the input / output formats for more info.

docs/en/guides/developer/working-with-json/json-load-data.md
DanRoscigno: ok, I will look into it.  Which of the three choices would you pick?

DanRoscigno: oh, I see above: use tuple.

Avogar: Yes, in this case when we 100% know that we always have 3 nested fields `method` and `path`, `version`, I would use Tuple

Avogar:
I would also added an example with this dataset: `https://datasets-documentation.s3.eu-west-3.amazonaws.com/github/github-2022.ndjson.gz` and enabled setting `input_format_json_read_objects_as_strings`

docs/en/guides/developer/working-with-json/json-load-data.md
Avogar: And actually, I think we can remove ephemeral column at all, just dont see shy it's really needed

docs/en/guides/developer/working-with-json/json-intro.md
```diff
@@ -1,18 +0,0 @@
----
-slug: /en/guides/developer/working-with-json/json-intro
-sidebar_label: Introduction 
-sidebar_position: 1
-description: Introduction to Handling JSON
----
-
-# Handling JSON
```
Avogar, line 8: Can we really want to just remove this intro?

docs/en/guides/developer/working-with-json/json-load-data.md
DanRoscigno: I think that the `@` symbol is causing a problem, I tried casting to DateTime, but that failed.  In fact, I cannot even select that column:
```sql
SELECT @timestamp FROM s3('https://datasets-documentation.s3.eu-west-3.amazonaws.com/http/documents-01.ndjson.gz', 
'JSONEachRow') LIMIT 1;
```

Fails with:
```
Expected one of: ALL, DISTINCT ON, DISTINCT, TOP, not empty list of expressions, list of expressions, list of elements, expression with optional alias, element of expression with optional alias, lambda expression, CAST operator, NOT, INTERVAL, CASE, DATE, TIMESTAMP, tuple, collection of literals, array, number, literal, NULL, Bool, true, false, string literal, asterisk, qualified asterisk, compound identifier, identifier, COLUMNS matcher, COLUMNS, qualified COLUMNS matcher, substitution, MySQL-style global variable
```

Avogar: You can put column name in quotes:
```sql
SELECT `@timestamp` FROM s3('https://datasets-documentation.s3.eu-west-3.amazonaws.com/http/documents-01.ndjson.gz', 'JSONEachRow') LIMIT 1;
```

DanRoscigno: :facepalm: I tried single quote, but not double!  Thanks.  `toDateTime` works fine 
```
SELECT toDateTime(`@timestamp`)
FROM s3('https://datasets-documentation.s3.eu-west-3.amazonaws.com/http/documents-01.ndjson.gz', 'JSONEachRow')
LIMIT 1

Query id: c27d4a3f-2097-4a7f-b3cb-d5ebe83871d5

┌─toDateTime(@timestamp)─┐
│    1998-04-30 19:30:17 │
└────────────────────────┘
```

docs/en/guides/developer/working-with-json/json-load-data.md
DanRoscigno: added describe

docs/en/guides/developer/working-with-json/json-semi-structured.md
DanRoscigno: agreed, I will update once I finish the json-load-data

docs/en/guides/developer/working-with-json/json-load-data.md
```diff
@@ -0,0 +1,147 @@
...
+# highlight-next-line
+   request Nested(method LowCardinality(String), path String, version LowCardinality(String)),
+   status       UInt16,
+   size         UInt32
+) ENGINE = MergeTree() ORDER BY (status, timestamp);
+```
+
+## Insert one row
+```sql
+INSERT INTO http (timestamp, clientip, request.method, request.path, request.version, status, size) SELECT
```
Avogar, line 70: I think using list of columns to insert is unnececcary here because we have all columns in the select query in right order

docs/en/guides/developer/working-with-json/json-intro.md
DanRoscigno: I will move the info into the other pages, I think the first doc that people see should tell them how to load data, not about all of the options.  At then end of loading data I will point them to the options, including the newer experimental choices.

docs/en/guides/developer/working-with-json/json-load-data.md
DanRoscigno: yes, I bet that was in there because of the ephemeral before, I will remove it

DanRoscigno: works fine, removed.

DanRoscigno:
I am going to switch to tuple now, thanks again @Avogar

DanRoscigno:
The load-data doc is updated for tuples, and additional info is added at end.  I am going to revert the changes to the semi-structured doc for now and pick that up next week when I have more time to dig into it.  Thanks again @Avogar !

docs/en/guides/developer/working-with-json/json-load-data.md
```diff
@@ -0,0 +1,172 @@
...
+- Import the dataset from S3
+
+## Examine the structure of the JSON file
+Examine the structure and one record from the log file in S3.  The `s3` function
+retrieves and decompresses the file and allows querying the file
+in S3 without loading it.
+
+This is what a row of the file contains:
+```json
+{"@timestamp":893964617,"clientip":"40.135.0.0","request":["GET", "/images/hm_bg.jpg", "HTTP/1.0"],"status":200,"size":24736}
```
Avogar, line 25: Actually data is a bit different:
```suggestion
{"@timestamp":893964617,"clientip":"40.135.0.0","request":{"method":"GET","path":"/images/hm_bg.jpg","version":"HTTP/1.0"},"status":200,"size":24736}
```

DanRoscigno:
> I would change it in a bit different way. Let's keep it as it was, but add `SETTINGS allow_experimental_object_type=1` and after that add same queries with `SETTINGS input_format_json_read_objects_as_strings=1` to show that we can load such dataset without using JSON object type.

moved to https://github.com/ClickHouse/clickhouse-docs/issues/659 I will get it done

PR #666, opened 2023-01-12
Simplify INSERT queries in json-load-data.md

Avogar:


docs/en/guides/developer/working-with-json/json-load-data.md
```diff
@@ -80,28 +80,21 @@ The `request` field from the JSON file will be stored as a tuple.
...
 ```sql
-INSERT INTO http SELECT
-# highlight-next-line
-    toDateTime(`@timestamp`) AS timestamp,
-    clientip,
-# highlight-next-line
-    (request['method'], request['path'], request['version']),
-    status,
-    size
+INSERT INTO http SELECT *
```
Avogar, line 97: In case of `SELECT * FROM ...` we will use structure from insertion table instead of schema inference, so no need to cast timestamp and select `method`,`path` and `version` explicitly

DanRoscigno (approved):
LGTM

PR #688, opened 2023-01-19
Update cache s3 multi region

DanRoscigno:
the cache system was changed, updating pages @e-mars

### DIFF
--- a/docs/en/integrations/data-ingestion/s3/configuring-s3-for-clickhouse-use.md
+++ b/docs/en/integrations/data-ingestion/s3/configuring-s3-for-clickhouse-use.md
@@ -156,10 +156,13 @@ vim /etc/clickhouse-server/config.d/storage_config.xml
         <access_key_id>ABC123</access_key_id>
         <secret_access_key>Abc+123</secret_access_key>
         <metadata_path>/var/lib/clickhouse/disks/s3_disk/</metadata_path>
-        <cache_enabled>true</cache_enabled>
-        <data_cache_enabled>true</data_cache_enabled>
-        <cache_path>/var/lib/clickhouse/disks/s3_disk/cache/</cache_path>
       </s3_disk>
+      <s3_cache>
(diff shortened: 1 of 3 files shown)


e-mars (changes requested):
Maybe in the note, add a link to https://clickhouse.com/docs/en/operations/storing-data/#using-local-cache since there was a change in 22.8 in cache configuration?

DanRoscigno:
> Maybe in the note, add a link to https://clickhouse.com/docs/en/operations/storing-data/#using-local-cache since there was a change in 22.8 in cache configuration?

Which note?

e-mars (approved):
discussed note in slack.

docs/en/integrations/data-ingestion/s3/configuring-s3-for-clickhouse-use.md
```diff
@@ -175,8 +178,7 @@ vim /etc/clickhouse-server/config.d/storage_config.xml
 ```
 
 :::note
-The tag `<s3_disk>` within the `<disks>` tag is an arbitrary label. This can be set to something else but the same label must be used in the `<disk>` tab under the `<policies>` tab to reference the disk.
-The `<metadata_path>` and `<cache_path>` are recommended to also include the name in the path to be able to identify the locations on disk.
+The tags `s3_disk` and `s3_cache` within the `<disks>` tag are arbitrary labels. These can be set to something else but the same label must be used in the `<disk>` tab under the `<policies>` tab to reference the disk.
 The `<S3_main>` tag is also arbitrary and is the name of the policy which will be used as the identifier storage target when creating resources in ClickHouse.
```
DanRoscigno, line 183:
```suggestion

The configuration shown above is for ClickHouse version 22.8 or higher, if you are using an older version please see the [storing data](/docs/en/operations/storing-data.md/#using-local-cache) docs.

```

PR #694, opened 2023-01-20
Integrate some tidbits from duplicate doc writing guide into this guide

rschu1ze:
(1) is an older and seemingly unmaintained version of (2). Let's not maintain two documents with the same content.

Some information existed in (1) but not in (2). This commit integrates it into (2). A separate PR will replace (1) by a (textual) pointer to (2).

EDIT: That separate PR is https://github.com/ClickHouse/ClickHouse/pull/45472

(1) https://github.com/ClickHouse/ClickHouse/blob/master/docs/tools/README.md
(2) https://github.com/ClickHouse/clickhouse-docs/blob/main/contrib-writing-guide.md

### DIFF
--- a/contrib-writing-guide.md
+++ b/contrib-writing-guide.md
@@ -23,9 +23,12 @@ Every page in the docs has an **Edit this page** link that opens the page in the
 
 Usually, these plugins provide a preview of how the markdown will render, and they catch basic errors like unclosed tags very early.
 
-## Building the docs
+## Building the docs locally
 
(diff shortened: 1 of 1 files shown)


DanRoscigno (changes requested):
Thanks for reminding me about this Robert!

rschu1ze:
@DanRoscigno Thanks for double-checking. Seems like I hit the "merge" button too quickly 😄 I pushed your suggestions here: #695

DanRoscigno:
Thanks @rschu1ze we crossed in the ether :)

PR #712, opened 2023-01-26
rename clickhouse-jdbc repo to clickhouse-java

mshustov:


zhicwu:
Looks good, thanks @mshustov! Will it be better to use [latest](https://github.com/ClickHouse/clickhouse-jdbc/releases/latest) instead of specific version?

By the way, ClickHouse/clickhouse-jdbc#1205 is for renaming branch from `master` to `main`. I'll submit a separate PR for the new repo name.

PR #715, opened 2023-01-27
Add docs for cascading materialized view

DanRoscigno:
@CamiloSierraH I will run through this and add output for the test queries, let me know if there are other docs that we should point to in the more info and any changes that you want.

image


rfraposa:
This is great @CamiloSierraH - thanks!

We could add a link to the on-demand MV course: https://learn.clickhouse.com/visitor_catalog_class/show/1043451/

PR #723, opened 2023-01-30
separate Cloud and Self-managed user management

DanRoscigno:
closes https://github.com/ClickHouse/product/issues/774


e-mars:
@DanRoscigno Recommend to add a note in the Users and Roles section near the top to state that it is either for Self-Managed or for Cloud and a link to the corresponding alternate page in case a google search takes them to the incorrect one for their use-case. 
Also maybe set the section name the same? one has `Defining SQL Users and Roles` and the other is just `Users and Roles`.
WDYT?

PR #725, opened 2023-01-30
Create a new guide for using TTL

rfraposa:
Including removing rows, removing columns, rolling up data, and an example of a hot/warm/cold architecture


DanRoscigno (approved):
This is great.  I added a note about config.d

PR #732, opened 2023-02-01
Added Lightweight Delete internals

davenger:
depends on https://github.com/ClickHouse/ClickHouse/pull/46015


DanRoscigno:
Thanks very much @davenger .  This is very clear information.  I would like to combine this page with the existing `sql-reference/statements/delete` page and then have the single page show in both the reference and in the guides (single-sourced, I can include it in the guides via an `import`).  This will help with maintenance going forward when there are changes as this is experimental.

What do you think?

davenger:
@DanRoscigno yep, this totally makes sense.

PR #734, opened 2023-02-02
ClickHouse-JS 0.0.12 update

slvrtrn:
Update the docs with an entry regarding the newly introduced `query_id`


docs/en/integrations/language-clients/nodejs.md
```diff
@@ -32,9 +32,9 @@ npm i @clickhouse/client
 
 ## Compatibility with ClickHouse
 
-| Client version | ClickHouse   |
-|----------------|--------------|
-| 0.0.1 - 0.0.11 | 22.8 - 22.11 |
+| Client version | ClickHouse  |
+|----------------|-------------|
+| 0.0.1 - 0.0.12 | 22.8 - 23.1 |
```
mshustov, line 37: Do we test old versions against 22.8? I don't think so. Let's add the actualized table. We can even remove old versions from the tablee

mshustov (approved):

PR #748, opened 2023-02-07
remove home and replace with "what is clickhouse?"

DanRoscigno:
- Removes /home from nav (renamed to _home so that we have it available for reference)
- Makes What is ClickHouse the landing page
- Updates redirects
- Moves About Us to top of nav
- Expands About Us

image


rfraposa:
Awesome - thanks!

PR #758, opened 2023-02-08
Fix gcs doc

kssenii:


docs/en/integrations/data-ingestion/s3/gcs-merge-tree.md
```diff
@@ -76,19 +70,19 @@ The example configuration highlighted below enables a 10Gi memory cache for the
...
-            </cache>
+            </gcs_cache>
 	    <!--highlight-end-->
         </disks>
         <policies>
             <gcs_main>
                 <volumes>
                     <main>
-                        <disk>gcs</disk>
+                        <disk>gcs_cache</disk>
```
kssenii, line 85: This place was wrong - it did not enable cache as stated “The example configuration highlighted below enables a 10Gi memory cache for the disk `gcs`.” . It made a user wonder why it does not work: https://clickhousedb.slack.com/archives/CU478UEQZ/p1675420392261419?thread_ts=1675350475.829829&cid=CU478UEQZ.

docs/en/integrations/data-ingestion/s3/gcs-merge-tree.md
```diff
@@ -40,12 +40,6 @@ This part of the configuration is shown in the highlighted section and specifies
                 <metadata_path>/var/lib/clickhouse/disks/gcs/</metadata_path>
 	    <!--highlight-end-->
             </gcs>
-	    <cache>
-                <type>cache</type>
-                <disk>gcs</disk>
-                <path>/var/lib/clickhouse/disks/gcs_cache/</path>
-                <max_size>10Gi</max_size>
-            </cache>
```
kssenii, line 48: I removed this because it does not make sense to have it here as `cache` is not added in `disk` section in `volumes`, then it is just redundant and confuses readers as they start to think that this will enable cache but it doesn't.

alexey-milovidov (approved):

docs/en/integrations/data-ingestion/s3/gcs-merge-tree.md
DanRoscigno: @kssenii I don't think gcs_cache belongs in the policies.

kssenii: what do you mean? Otherwise cache will not be enabled.

PR #834, opened 2023-02-26
Adding icons for recently added integrations

tbragin:
Adding icons for integrations that didn't have them, where I could find them. I tested locally with Chrome browser and it looked ok :-)


tbragin:
Per @DanRoscigno going to merge since tests passed.

PR #872, opened 2023-03-05
Schema visualization and database doc generation

igor-suhorukov:


rfraposa:
SchemaSpy looks pretty cool! I've added it to the /docs/en/integrations page, which seems like a more appropriate spot.

PR #890, opened 2023-03-09
standardize admonitions

DanRoscigno:
Many of the admonitions are severe, where a tip, note, info, or important is appropriate.  Lets use warning when there is risk of data loss or system crash.


rfraposa:
I might revert this PR....some of these files don't exist anymore.

Moving forward, you should make changes like this to the `new-nav` branch.

PR #895, opened 2023-03-10
Update changelog.md

tbragin:
Removing "backups optimization" from Dec 20, since we ended up shipping it on March 9 instead.

@DanRoscigno


DanRoscigno (approved):
Thanks Tanya

PR #927, opened 2023-03-22
[Feature] Add zing docs

viduni94:


mshustov:
You might want to add zing to https://github.com/ClickHouse/clickhouse-docs/blob/main/docs/en/integrations/index.mdx to render on https://clickhouse.com/docs/en/integrations page as well

docusaurus.config.js
rfraposa, line 1: Why do you need a redirect? Those are for old pages that get moved....but this appears to be a new docs page

viduni94: Got it. I removed the redirect @rfraposa 
Thank you

PR #933, opened 2023-03-23
Remove folder links in nav

rfraposa:


rfraposa:
I'm in catch-22 hell...I need to force merge this branch so that the corresponding changes in ClickHouse/ClickHouse can be merged, or it will never build properly.

PR #938, opened 2023-03-24
[KB] Importing and querying JSON array objects

nellicus:
KB article on Importing and querying JSON array objects

### DIFF
--- a/knowledgebase/importing_and_working_with_JSON_array_objects.md
+++ b/knowledgebase/importing_and_working_with_JSON_array_objects.md
@@ -1,0 +1,296 @@
+---
+date: 2023-03-24
+---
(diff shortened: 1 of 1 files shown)


nellicus:
unsure why check fails, this is all I can see in the job execution log

```
The self-hosted runner: i-02b944540ee5063ed lost communication with the server. Verify the machine is running and has a healthy network connection. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.
```

rfraposa:
I can't figure out the checks either - I think there is a bug in our CI. I'll merge anyway...this looks great! Thank you for contributing.

PR #995, opened 2023-04-06
Adding new brand colour theme

gjones:
### Summary 
This PR adds the colour theme of the new brand to the ClickHouse docs along with some minor changes to the navigation.

#### Screenshots
CleanShot 2023-04-06 at 07 10 01@2x
CleanShot 2023-04-06 at 07 09 36@2x


DanRoscigno:
is there a list of issues that the updates are solving?

gjones:
No, just changing colours and centring the navigation to align more with the website.

vineethasok:
@gjones Are we going to remove the light mode?
Currently we are only using dark mode for the website, that's why I asked

vineethasok:
@gjones Need to update the favicon too

gjones:
> @gjones Are we going to remove the light mode?

Nope, users should be able to choose their preferred colour scheme here. The control plane will also have multiple themes.

For now I've changed the default theme to dark, however I left `preferssystemtheme` to true which means most people will still get light. I'll probably turn that off and default to dark, then if they want light, they can hit the button, thoughts? 

Yep I'll update the favicon

PR #1051, opened 2023-04-19
Update FAQ to reflect lightweight-deletes as GA

tylerhannan:
Not sure if it is the best approach to reference enabling for prior versions. Just a proposal. Please feel free to tweak appropriately.


DanRoscigno (approved):
Thanks Tyler.  This is good.  We do need to figure out how to deal with versioned docs, I think it is going to be considered for next quarter

PR #1335, opened 2023-07-10
Update to ClickPipes guide

Ryado:


DanRoscigno:
@Ryado I don't think that we should mix the terms "Beta" and "Private Preview" in one page.  Aren't they different?  Maybe "Private Beta"?

Ryado:
`Beta` is the feature readiness and `Private Preview` is the model the feature is released under

PR #1336, opened 2023-07-10
Adding Dictionary String type KB article

jaijhala:
Pull Request for KB article on how to add dictionary with string key/values.

### DIFF
--- a/knowledgebase/dictionary_using_strings.md
+++ b/knowledgebase/dictionary_using_strings.md
@@ -1,0 +1,65 @@
+---
+title: How to create a dictionary with String type keys and values
+date: 2023-07-10
+---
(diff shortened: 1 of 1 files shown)


DanRoscigno:
What do you think about an example query that uses a dictionary function:
```sql
SELECT dictGet('db1.table1_dict', 'name', 'me@domain.com')
```
```response
┌─dictGet('db1.table1_dict', 'name', 'me@domain.com')─┐
│ me                                                  │
└─────────────────────────────────────────────────────┘
```

jaijhala (commented):
Updated dicGet example.

DanRoscigno (approved):

PR #1453, opened 2023-08-18
Put SELECT command into codeblock.

johnnymatthews:
Judging from the rest of the tutorial, SQL commands should be contained within ` codeblocks. In [step 6](https://clickhouse.com/docs/en/getting-started/quick-start#6-query-your-new-table) the SELECT command is used but isn't in a codeblock. This PR fixes that.


johnnymatthews:
Never seen that CLA Assistant bot before. That's a nice touch!

alexey-milovidov (approved):

PR #1538, opened 2023-09-22
Initial Snowflake migration docs

justindeguzman:
This is a simplified doc of the migration section from this blog post:
https://clickhouse.com/blog/clickhouse-vs-snowflake-for-real-time-analytics-comparison-migration-guide#migrating-data


kellytoole (approved):
This looks great! Thank you so much again

tbragin:
@gingerwizard any reviews? what are the next steps here? cc @Ryado

docs/en/migrations/snowflake.md
```diff
@@ -0,0 +1,110 @@
...
+URL='s3://mybucket/mydataset'
+CREDENTIALS=(AWS_KEY_ID='<key>' AWS_SECRET_KEY='<secret>')
+FILE_FORMAT = my_parquet_format;
+
+-- Apply "mydataset" prefix to all files and specify a max file size of 150mb
+-- The `header=true` parameter is required to get column names
+COPY INTO @external_stage/mydataset from mydataset max_file_size=157286400 header=true;
+```
+
+For a dataset around 5TB of data with a maximum file size of 150MB, and using a 2X-Large Snowflake warehouse located in the same AWS `us-east-1` region, copying data to the S3 bucket will take around 30 minutes.
```
gingerwizard, line 50: Not verified this, i assume you lifted from the blog and sounds roughly right. Users should only pay for warehouse compute if its in the same region also.

docs/en/migrations/snowflake.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+We can then use the `INSERT INTO SELECT` command to insert the data from S3 into a ClickHouse table:
+
+```sql
+INSERT INTO mydataset
+SELECT
+	timestamp,
+	some_text,
+	JSONExtract(
```
gingerwizard, line 79: last time i checked they do export OBJECT as JSON string..this is subject to change. i raised a question and they planned to possibly change. Maybe add note?

justindeguzman:
> LGTM. Maybe note the string export of JSON is subject to change and users should do a DESCRIBE s3() toi check the schema

Just to keep it simple, I don't think we should add this info unless Snowflake's behavior actually changes.

Thanks for reviewing @gingerwizard!

PR #1547, opened 2023-09-26
Initial BigQuery migration guide

justindeguzman:
This is a simplified doc of the migration section from this blog post:
https://clickhouse.com/blog/clickhouse-bigquery-migrating-data-for-realtime-queries#exporting-data-to-gcs

### DIFF
--- a/docs/en/migrations/bigquery.md
+++ b/docs/en/migrations/bigquery.md
@@ -103,4 +103,8 @@ SELECT * FROM mytable limit 10;
 
 To export more BigQuery tables, simply redo the steps above for each additional table.
 
+## Further reading and support
+
+In addition to this guide, we also recommend reading our blog post that shows [how to use ClickHouse to speed up BigQuery](https://clickhouse.com/blog/clickhouse-bigquery-migrating-data-for-realtime-queries).
+
 If you are having issues transferring data from BigQuery to ClickHouse, please feel free to contact us at support@clickhouse.com.
--- a/docs/en/migrations/bigquery.md
+++ b/docs/en/migrations/bigquery.md
@@ -1,0 +1,106 @@
+---
+sidebar_label: BigQuery
+sidebar_position: 20
+slug: /en/migrations/bigquery
+description: Migrating from BigQuery to ClickHouse
+keywords: [migrate, migration, migrating, data, etl, elt, bigquery]
+---
+
+# Migrating from BigQuery to ClickHouse
+
+This guide shows how to migrate data from [BigQuery](https://cloud.google.com/bigquery) to ClickHouse.
+
+In this guide, we first export a table to [Google's object store (GCS)](https://cloud.google.com/storage) and then import that data into [ClickHouse Cloud](https://clickhouse.com/cloud). These steps need to be repeated for each table you wish to export from BigQuery to ClickHouse.
+
+## How long will exporting data to ClickHouse take?
+
+Exporting data from BigQuery to ClickHouse is dependent on the size of your dataset. As a comparison, it takes about an hour to export the [4TB public Ethereum dataset](https://cloud.google.com/blog/products/data-analytics/ethereum-bigquery-public-dataset-smart-contract-analytics) from BigQuery to ClickHouse using this guide.
+
+| Table                                                                                             | Rows          | Files Exported | Data Size | BigQuery Export | Slot Time       | ClickHouse Import |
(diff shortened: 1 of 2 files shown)


robot-ch-test-poll4:
*This is an automated comment for commit b29a009dd027025e6440ac27661ba3c392dbf794 with description of existing statuses. It's updated for the latest CI running*

[✅ Click here](https://s3.amazonaws.com/clickhouse-test-reports/1547/b29a009dd027025e6440ac27661ba3c392dbf794/ci_running.html) to open a full report in a separate page

<details><summary>Successful checks</summary>
<table>
<thead><tr><th>Check name</th><th>Description</th><th>Status</th></tr></thead>
<tbody>
<tr><td>Docs Check</td><td>Builds and tests the documentation</td><td>✅ success</td></tr>
<tr><td>Vercel</td><td>There's no description for the check yet, please add it to tests/ci/ci_config.py:CHECK_DESCRIPTIONS</td><td>✅ success</td></tr>
<tbody>
</table>
</details>

tbragin:
@kellytoole @gingerwizard could you review this PR when you get a chance?

docs/en/migrations/bigquery.md
```diff
@@ -0,0 +1,110 @@
...
+);
+```
+
+In the above query, we export our BigQuery table to the [Parquet data format](https://parquet.apache.org/). We also have a `*` character in our `uri` parameter. This ensures the output is sharded into multiple files, with a numerically increasing suffix, should the export exceed 1GB of data.
+
+This approach has a number of advantages:
+
+- Google allows up to 50TB per day to be exported to GCS for free. Users only pay for GCS storage.
+- Exports produce multiple files automatically, limiting each to a maximum of 1GB of table data. This is beneficial to ClickHouse since it allows imports to be parallelized.
+- Parquet, as a column-oriented format, represents a better interchange format since it is inherently compressed and faster for BigQuery to export and ClickHouse to query
```
gingerwizard, line 49: For larger exports the following can be used:

```
DECLARE export_path string;
DECLARE n int64;
DECLARE i INT64;
SET i = 0;
SET n = 100;


WHILE i < n DO
  SET export_path = CONCAT('gs://mybucket/mytable/', i,'-*.parquet');
  EXPORT DATA
    OPTIONS (
    uri = (export_path),
    format = 'PARQUET',
    overwrite = true
    )
  AS (
    SELECT *
      FROM mytable
    WHERE export_id = i
  );
  SET i = i + 1;
END WHILE;
```

docs/en/migrations/bigquery.md
```diff
@@ -0,0 +1,110 @@
...
+    '<ACCESS_ID>',
+    '<SECRET>'
+);
+```
+
+The `ACCESS_ID` and `SECRET` used in the above query is your [HMAC key](https://cloud.google.com/storage/docs/authentication/hmackeys) associated with your GCS bucket.
+
+:::note Use `ifNull` when exporting nullable columns
+In the above query, we use the [`ifNull` function](/docs/en/sql-reference/functions/functions-for-nulls#ifnull) with the `some_text` column to insert data into our ClickHouse table with a default value. You can also make your columns in ClickHouse [`Nullable`](/docs/en/sql-reference/data-types/nullable), but this is not recommended as it may affect negatively performance.
+:::
```
gingerwizard, line 94: Might be worth bringing the following to the users attention:

input_format_null_as_default=1 
input_format_parquet_import_nested=1

docs/en/migrations/bigquery.md
```diff
@@ -0,0 +1,110 @@
...
+
+```sql
+SELECT * FROM mytable limit 10;
+```
+
+To export more BigQuery tables, simply redo the steps above for each additional table.
+
+## Further reading and support
+
+In addition to this guide, we also recommend reading our blog post that shows [how to use ClickHouse to speed up BigQuery](https://clickhouse.com/blog/clickhouse-bigquery-migrating-data-for-realtime-queries).
```
gingerwizard, line 108: and how possibly incremental import can be handled

gingerwizard (commented):
initial feedback

docs/en/migrations/bigquery.md
justindeguzman: In this query, is there an optimal `n` we should suggest, like if 1TB split it up into `n` files per `x` GB?

Also, can we use this query even for small datasets? I'd like to simplify the guide so just one query for both small and large imports would be easier in my opinion for the user.

gingerwizard: this is really when you have over 5 billion rows. Maybe total rows/5billion = N

PR #1592, opened 2023-10-17
Remove region limitation for ClickPipes

jkaflik:


johnnymatthews (approved):
Looks fine from my perspective, but I'll let @Ryado confirm whether the region limitation has been lifted or not.

PR #1870, opened 2024-01-22
fixing typos (remove SVG from name)

Ryado:


johnnymatthews (approved):
Ooph, this was probably me. Nice catch though, thanks!

PR #1874, opened 2024-01-24
Update aws private link INSTANCE_ID jq command

allegrinisante:
In case of an organization with multiple services, the current command will populate INSTANCE_ID with the first service_id.
With this change, the filter will match also on the desired instance name.

### DIFF
--- a/docs/en/cloud/security/aws-privatelink.md
+++ b/docs/en/cloud/security/aws-privatelink.md
@@ -36,7 +36,7 @@ ORG_ID=<Your ClickHouse organization ID>
 SERVICE_NAME=<Your ClickHouse service name>
 ```
 
-Get the desired instance ID by filtering on region, provider and service name:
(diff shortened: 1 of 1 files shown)


docs/en/cloud/security/aws-privatelink.md
```diff
@@ -33,15 +33,15 @@ PROVIDER=aws
 KEY_ID=<Your key ID>
 KEY_SECRET=<Your key secret>
 ORG_ID=<Your ClickHouse organization ID>
+SERVICE_NAME=<Your ClickHouse service name>
 ```
 
-Next, get an instance ID from your region:
+Get the desired instance ID by filtering on region, provider and service name:
```
johnnymatthews, line 39:
```suggestion
Get the desired instance ID by filtering by region, provider, and service name:
```

johnnymatthews (approved):
Looks good!

PR #1864, opened 2024-01-19
Updating cloud access management docs

leticiawebb:
Consolidated user permissions from the Manage section and users-and-roles from the Security section into an overall Cloud Access Management doc with illustrations.

PR #2045, opened 2024-02-12
fix broken link

mshustov:
broken by https://github.com/ClickHouse/clickhouse-docs/pull/1864
needs to be tested.

I have a few questions:
- does redirect plugin work locally? it doesn't seem to work for me
- don't we have automation to detect broken links?

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -697,6 +697,10 @@ const config = {
 						to: '/en/operations/access-rights',
 					},
 					{
+						from: '/en/cloud/users-and-roles',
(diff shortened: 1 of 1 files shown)


pjhampton (approved):

johnnymatthews:
> does redirect plugin work locally? it doesn't seem to work for me

No. It does on standard Docusaurus sites, but not our because we've borked it so much.

> don't we have automation to detect broken links?

We do, but not for the sidebar apparently. Something we should fix. I imagine a regular Docusaurus site has this feature built in.

johnnymatthews (approved):

PR #2046, opened 2024-02-13
[ClickPipes] De-emphasise Kafka in docs

pjhampton:
We're soon going to move beyond Kafka and Managed Kafka Providers. 
This is the first step to deemphasise Kafka as the only source and just make instructions more generic.

### DIFF
--- a/docs/en/integrations/data-ingestion/clickpipes/index.md
+++ b/docs/en/integrations/data-ingestion/clickpipes/index.md
@@ -11,14 +11,13 @@ import AzureEventHubsSVG from "../../images/logos/azure_event_hubs.svg";
 import UpstashSVG from "../../images/logos/upstash.svg";
 import WarpStreamSVG from "../../images/logos/warpstream.svg";
 
-# Integrating Kafka with ClickHouse Cloud
+# Integrating with ClickHouse Cloud
 
 ## Introduction
 
-[ClickPipes](https://clickhouse.com/cloud/clickpipes) is a managed integration platform that makes ingesting data from a diverse set of sources as simple as clicking a few buttons. Designed for the most demanding workloads, ClickPipes's robust and scalable architecture ensures consistent performance and reliability. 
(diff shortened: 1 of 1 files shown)


docs/en/integrations/data-ingestion/clickpipes/index.md
```diff
@@ -114,10 +102,10 @@ The supported formats are:
 
 | Format                                                                                    | Support     |
```
kschlarman, line 103: @Ryado Would it make sense to expand this graph to show which formats are supported for which providers? For example, Avro will be supported for Kafka, but not S3. 
(Sorry, outside the scope of this PR but relevant to how this page becomes more generic).

docs/en/integrations/data-ingestion/clickpipes/index.md
```diff
@@ -114,10 +102,10 @@ The supported formats are:
...
 |-------------------------------------------------------------------------------------------|-------------|
-| [JSON](../../../interfaces/formats.md/#json)                                               | ✔           | 
+| [JSON](../../../interfaces/formats.md/#json)                                               | ✔           |
 | [AvroConfluent](../../../interfaces/formats.md/#data-format-avro-confluent)                |*Coming Soon*|
-| [TabSeparated](../../../interfaces/formats.md/#tabseparated)                               |*Coming Soon*| 
-| [CSV](../../../interfaces/formats.md/#csv)                                                 |*Coming Soon*| 
+| [TabSeparated](../../../interfaces/formats.md/#tabseparated)                               |*Coming Soon*|
+| [CSV](../../../interfaces/formats.md/#csv)                                                 |*Coming Soon*|
 
 ## Supported data types
```
kschlarman, line 110: I think this might need to be forked for cases where standard JSON is not the source? For S3, we'll support anything clickhouse itself supports. I'm guessing Avro gets more complicated cc @genzgd. 
(Sorry, outside the scope of this PR but relevant to how this page becomes more generic).

kschlarman (approved):
Seems like a step in the right direction, but I think making supported formats/data types sections generic might be hard.

PR #2057, opened 2024-02-17
Update changelog.md

melvynator:
Add some details about a few settings.


tbragin (approved):
Content is good but I suggest small stylistic tweaks cc @justindeguzman 

* `use_mysql_types_in_show_columns` is removed, it will be automatically enabled once you connect through the MySQL interface
* `async_insert_max_data_size` now has the default value of 10 MiB

justindeguzman (approved):

PR #2067, opened 2024-02-20
Update json.md

Blargian:
Correct type of 'response' field to 'request' field:

Screenshot from 2024-02-20 19-02-49

In the image above the field which contains the nested JSON is called 'request', not 'response'.


johnnymatthews (approved):
Nice catch! Thanks!

PR #2102, opened 2024-03-02
Add legacy approach to import GPG keys

peter279k:
# Changed log

- Add the alternative way about importing the GPG keys with `gpg` command.
- It will be used when the GPG versions are `1.x` or `2.0.x`.
- This approach is legacy because the GPG key will be stored/imported in the global primary keyring.


peter279k:
Closed because it's not necessary now since the [#61252](https://github.com/ClickHouse/ClickHouse/pull/61252) is merged.

PR #2127, opened 2024-03-11
Remove links to `clickhouse-copier`

nikitamikhaylov:
`clickhouse-copier` is removed from the bundle.


Vukiz:
@nickpresta https://github.com/ClickHouse/copier it is obsolete

PR #2138, opened 2024-03-16
Update the JS client docs - 0.3.0

slvrtrn:
* Update the docs according to the [upcoming 0.3.0 release](https://github.com/ClickHouse/clickhouse-js/blob/main/CHANGELOG.md#030-nodejs-only).
* Fixed dead links
* Rewritten outdated sections
* Rewritten certain sections for clarity
* Added more information about the ResultSet
* Added the [examples overview](https://github.com/ClickHouse/clickhouse-js/blob/main/examples/README.md#overview) link
* Added the "Contact Us" section
* Removed "Endless streams" examples, as using async inserts will be a much safer option for the same use cases
* Highlighted and clarified certain features (for example, JSON _ClickHouse format_ cannot be streamed, but JSONEachRow can be; it can lead to confusion if one is unfamiliar with CH formats and could conclude that the client does not support streaming JSON data at all, while it does have this feature).
* Supported Node.js versions table is reintroduced
* Update the Keep-Alive configuration section with the latest insights.

### DIFF
--- a/docs/en/integrations/language-clients/js.md
+++ b/docs/en/integrations/language-clients/js.md
@@ -36,7 +36,6 @@ Current Node.js versions support:
 | 18.x            | ✔           |
 | 16.x            | Best effort |
 
-
 ## Environment requirements (Web)
 
 Web version of the client is officially tested with the latest Chrome/Firefox browsers and can be used as a dependency in, for example, React/Vue/Angular applications, or CloudFlare workers.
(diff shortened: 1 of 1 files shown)


docs/en/integrations/language-clients/js.md
```diff
@@ -47,17 +57,26 @@ npm i @clickhouse/client-web
 
 ## Compatibility with ClickHouse
 
-| Client version | ClickHouse   |
-|----------------|--------------|
-| 0.2.9          | 22.8 - 23.12 |
+| Client version | ClickHouse  |
+|----------------|-------------|
+| 0.3.0          | 22.8 - 24.2 |
```
mshustov, line 62: Do we run tests against `22.8`? Why? `23.3` is the latest supported https://github.com/ClickHouse/ClickHouse/blob/master/SECURITY.md

slvrtrn: Update this part; thanks for pointing it out.

PR #2154, opened 2024-03-21
Add docs for clickhouse-static-files-disk-uploader

justindeguzman:


kssenii (approved):

alexey-milovidov:
Add a link to https://github.com/ClickHouse/web-tables-demo as an example.

docs/en/tools-and-utilities/static-files-disk-uploader.md
```diff
@@ -0,0 +1,82 @@
...
+|`--url [url]`|Web server URL for `test` mode|
+|`--output-dir [dir]`|Directory to output files in `non-test` mode|
+
+## Retrieve metadata path for the specified table
+
+When using `clickhouse-static-files-disk-uploader`, you will need to obtain the metadata path for your desired table.
+
+1. Run the following query specifying your target table and database:
+
+<br />
```
johnnymatthews, line 36: Are these `<br />` elements needed? If we want more padding between list elements and codeblocks, we should make changes to the CSS rather than hardcoding breaks like this.

docs/en/tools-and-utilities/static-files-disk-uploader.md
```diff
@@ -0,0 +1,82 @@
...
+
+```sql
+SELECT data_paths
+  FROM system.tables
+  WHERE name = 'mytable' AND database = 'default';
+```
```
johnnymatthews, line 42: This might help any padding/margin issues:

````suggestion
    ```sql
    SELECT data_paths
      FROM system.tables
      WHERE name = 'mytable' AND database = 'default';
    ```
````

docs/en/tools-and-utilities/static-files-disk-uploader.md
justindeguzman: We should–just for now since I wanted to commit this quickly the `<br/>` is temporarily to make it formatted nicely.

docs/en/tools-and-utilities/static-files-disk-uploader.md
justindeguzman: Does the padding look incorrect? It looks OK in my local environment.

johnnymatthews: It might be fine, I didn't actually look at a preview. But this suggestion was to remove the need for the `<br>` tags.

Issue #2199, opened 2024-04-05
Latex/Katex equations don't render


Blargian:
On the [Windows Functions page](https://clickhouse.com/docs/en/sql-reference/window-functions) it can be seen below that the latex equation is not rendering correctly. I noticed this issue on another page as well. (can't find it now because I fixed it by not using LaTex in the .md file).

![windows_functions_latex_display_issue](https://github.com/ClickHouse/clickhouse-docs/assets/41984034/d42749f4-428e-4238-a44d-efbb0c393f29)

PR #2200, opened 2024-04-05
Fix for broken LaTex rendering

Blargian:
Opening this PR as a potential fix to #2199.

**Original issue**: 

Latex equations do not render correctly:

windows_functions_latex_display_issue

Initially I thought it was a syntax issue with the LaTex in the .md file but after many unsuccessful attempts I concluded that LaTex/KaTex is just not supported in the docs, yet. (it doesn't seem commonly used across the docs site in any case which is why we didn't pick it up). 

**Changes made**:

I added in the plugins required for KaTex rendering as per the docusaurus documentation for version 2:
[docusuarus version 2 - adding math support](https://docusaurus.io/docs/2.x/markdown-features/math-equations)

**Result**:

LaTex now renders correctly:

windows_functions_latex_display_after_fix

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -1,5 +1,7 @@
 const darkTheme = require('prism-react-renderer/themes/vsDark')
 const path = require("path")
+const math = require('remark-math');
+const katex = require('rehype-katex');
 
 /** @type {import('@docusaurus/types').Config} */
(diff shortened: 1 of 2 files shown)


justindeguzman:
Thank you Shaun!

PR #2208, opened 2024-04-09
Prettify these query results in AWS S3 doc

peter279k:
# Changed log

- As title, I think it's great to make these query results pretty format.


pjhampton (commented):
Thanks for keeping these tables looking great, @peter279k!

I noticed in the deployment that there is another table that could do with updating in this file. Would you mind updating this one too?

Screenshot 2024-04-10 at 16 50 52

peter279k:
@pjhampton, I've updated and thanks for your reply.

pjhampton (approved):
Thank you for your contribution :-)

PR #2222, opened 2024-04-16
add version info to cloud page

mneedham:


tbragin:
Thank you!

krithika-ch (changes requested):
Added a mention of the "Fast Release Channel"

tbragin:
Thank you! Let's not lose the other edit - I suggest.

> Clickhouse Cloud continuously upgrades your service to a newer version. After publishing a core database version in the open source, we do additional validation in our cloud staging environment - a process that typically takes 6-8 weeks - before rollout to production. The rollout is phased out by cloud service provider, service type, and region.
> 
> We offer a "Fast" Release Channel for subscribing to updates ahead of the regular release schedule. Access to early upgrades is recommended only for non-production environments and can be requested by logging a support ticket.
> 
> If you rely on functionality in the earlier version, in some cases you can fall back to this previous behavior using the compatibility setting for your service.

PR #2271, opened 2024-05-01
Add "cloud only" badge

thevar1able:
I'll need this badge later in JWT docs.


justindeguzman:
@thevar1able there's actually a <CloudAvailableBadge/> component that we can use.

PR #2280, opened 2024-05-03
Remove "Coming soon" from Kafka streaming TSV support

justindeguzman:
https://clickhouse-inc.slack.com/archives/C03A9B0H05T/p1714745825524999


genzgd:
I don't think CSV is coming any sooner?  Do we want to remove that too?

pjhampton (approved):

PR #2314, opened 2024-05-14
azure private link documentation

tsolodov:
This is a draft of public document.


docs/en/cloud/security/azure-privatelink.md
```diff
@@ -153,31 +156,31 @@ resource "azurerm_private_endpoint" "example_clickhouse_cloud" {
...
 
-Under properties find `resourceGuid` field and copy this value:
+Under properties, find `resourceGuid` field and copy this value:
 
 ![PE GUID](@site/docs/en/cloud/security/images/azure-pe-resource-guid.png)
 
 ## Setting up DNS for Private Link
 
-You need will need to create Private DNS zone ${location_code}.privatelink.azure.clickhouse.cloud and attach it to VNET to access resources via Private Link.
+You need will need to create a Private DNS zone and attach it to your VNET to access resources via Private Link.
```
tsolodov, line 179: It's not obvious that customer need to create ${region_code}.privatelink.azure.clickhouse.cloud zone. Let's add it somewhere ?

docs/en/cloud/security/azure-privatelink.md
```diff
@@ -0,0 +1,486 @@
...
+
+```bash
+cat <<EOF | tee pl_config_org.json
+{
+  "privateEndpoints": {
+    "add": [
+      {
+        "cloudProvider": "azure",
+        "id": "${ENDPOINT_ID:?}",
+        "description": "Azure private endpoint"
```
tsolodov, line 286: "region": is missing

tsolodov, line 286:
```suggestion
        "description": "Azure private endpoint",
        "region": "${REGION:?}"
```

docs/en/cloud/security/azure-privatelink.md
justindeguzman: Should I add back the following?

>You need will need to create Private DNS zone `${location_code}.privatelink.azure.clickhouse.cloud` and attach it to VNET to access resources via Private Link.

And is it `location_code` or `region_code` (you referenced both)? Sorry, I have not used Azure before so I am unfamiliar with what this URL would be used for. Maybe adding a short description would be helpful.

tsolodov: In azure region called as "location", that's why I put `location_code` here. I am ok with both options, not sure what is less confusing thought.

justindeguzman (approved):

PR #2349, opened 2024-05-24
May 2024 security updates

leticiawebb:
There are a few changes in here:
- Restructuring the cloud security section of docs for better flow.
- Replacing our security best practices page with a more thorough security shared responsibility page.
- Fixing slugs, links, and images

Where I updated slugs, I also searched the code base to find instances of the old slug, replaced them, and tested. Let me know if I missed any.


mzitnik:
@leticiawebb From the partial pages (integration) LGTM, but the rest is DP

PR #2387, opened 2024-06-07
Add new ClickPipes NAT IP for productions regions

jkaflik:
Add an extra NAT IP that will be used in the future.


kitop (approved):
LGTM - IPs match the ones in https://github.com/ClickHouse/data-plane-infrastructure/pull/4717#issuecomment-2154635214

PR #2397, opened 2024-06-12
Update kinesis secure role setup to include ListStreams as a separate section of the policy with an allow all

tpanetti:
ListStreams unfortunately does not work without a `*` on the streams ([aws docs](https://docs.aws.amazon.com/streams/latest/dev/controlling-access.html#kinesis-using-iam-examples)). Adding this to our iam policy docs to avoid future confusion for customers.


tpanetti:
> Can we avoid the ListStreams call in some way if we know we have access to the configured stream?

We could allow direct entry of the arn for the stream they want to sample from. That would only require some minor changes to the api and ui. Might be a bit odd adding that for Kinesis and not Kafka though and might require some more thinking from the UI side.

PR #2412, opened 2024-06-18
Update the CHJS docs for 1.2.0

slvrtrn:
Added the docs about the HTTP agent override.

### DIFF
--- a/docs/en/integrations/language-clients/js.md
+++ b/docs/en/integrations/language-clients/js.md
@@ -1194,7 +1194,7 @@ const client = createClient({
 This is an experimental feature that may change in backwards-incompatible ways in the future releases. The default implementation and settings the client provides should be sufficient for most use cases. Use this feature only if you are sure that you need it.
 :::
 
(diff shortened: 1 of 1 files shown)


docs/en/integrations/language-clients/js.md
```diff
@@ -1186,6 +1188,86 @@ const client = createClient({
 })
 ```
 
+### Custom HTTP/HTTPS agent (experimental, Node.js only)
+
+:::warning
+This is an experimental feature that may change in backwards-incompatible ways in the future releases. The default implementation and settings the client provides should be sufficient for most use cases. Use this feature only if you are sure that you need it.
+:::
+
+By default, the client will configure the underlying HTTP(s) agent using the settings provided in the client configuration (such as `max_open_connections,` `keep_alive.enabled`, `tls`), which will handle the connections to the ClickHouse server. Additionally, if TLS certificates are used, the underlying agent will be configured with the necessary certificates, and the correct TLS auth headers will be enforced.
```
SpencerTorres, line 1197: Golden docs. Only one nit, but everything else looks good 👌 

```suggestion
By default, the client will configure the underlying HTTP(s) agent using the settings provided in the client configuration (such as `max_open_connections`, `keep_alive.enabled`, `tls`), which will handle the connections to the ClickHouse server. Additionally, if TLS certificates are used, the underlying agent will be configured with the necessary certificates, and the correct TLS auth headers will be enforced.
```

PR #2436, opened 2024-06-28
Update ClickHouse Keeper page

Blargian:
The settings on the [Clickhouse Keeper](https://clickhouse.com/docs/en/guides/sre/keeper/clickhouse-keeper) page are quite difficult to read. I have placed them into table form (like they are further down on this page). Makes it a little easier to read:

image


docs/en/guides/sre/keeper/index.md
```diff
@@ -33,57 +33,54 @@ External integrations are not supported.
...
-- `force_sync` — Call `fsync` on each write to coordination log (default: true).
-- `quorum_reads` — Execute read requests as writes through whole RAFT consensus with similar speed (default: false).
-- `raft_logs_level` — Text logging level about coordination (trace, debug, and so on) (default: system default).
-- `auto_forwarding` — Allow to forward write requests from followers to the leader (default: true).
-- `shutdown_timeout` — Wait to finish internal connections and shutdown (ms) (default: 5000).
-- `startup_timeout` — If the server doesn't connect to other quorum participants in the specified timeout it will terminate (ms) (default: 30000).
-- `four_letter_word_white_list` — White list of 4lw commands (default: `conf,cons,crst,envi,ruok,srst,srvr,stat,wchs,dirs,mntr,isro,rcvr,apiv,csnp,lgif,rqld,ydld`).
-- `async_replication` - Enable async replication. All write and read guarantees are preserved while better performance is achieved. Settings is disabled by default to not break backwards compatibility (default: false)
```
SpencerTorres, line 82: It looks like there's 5 entries below `four_letter_word_white_list` that are no longer present in the new table?
Do we still need these entries?

docs/en/guides/sre/keeper/index.md
```diff
@@ -21,9 +21,9 @@ ClickHouse Keeper provides the coordination system for data [replication](/docs/
...
 
-By default, ClickHouse Keeper provides the same guarantees as ZooKeeper (linearizable writes, non-linearizable reads). It has a compatible client-server protocol, so any standard ZooKeeper client can be used to interact with ClickHouse Keeper. Snapshots and logs have an incompatible format with ZooKeeper, but the `clickhouse-keeper-converter` tool enables the conversion of ZooKeeper data to ClickHouse Keeper snapshots. The interserver protocol in ClickHouse Keeper is also incompatible with ZooKeeper so a mixed ZooKeeper / ClickHouse Keeper cluster is impossible.
+By default, ClickHouse Keeper provides the same guarantees as ZooKeeper - linearizable writes and non-linearizable reads. It has a compatible client-server protocol, so any standard ZooKeeper client can be used to interact with ClickHouse Keeper. Snapshots and logs have an incompatible format with ZooKeeper, but the `clickhouse-keeper-converter` tool enables the conversion of ZooKeeper data to ClickHouse Keeper snapshots. The interserver protocol in ClickHouse Keeper is also incompatible with ZooKeeper so a mixed ZooKeeper / ClickHouse Keeper cluster is impossible.
```
SpencerTorres, line 26:
```suggestion
By default, ClickHouse Keeper provides the same guarantees as ZooKeeper: linearizable writes and non-linearizable reads. It has a compatible client-server protocol, so any standard ZooKeeper client can be used to interact with ClickHouse Keeper. Snapshots and logs have an incompatible format with ZooKeeper, but the `clickhouse-keeper-converter` tool enables the conversion of ZooKeeper data to ClickHouse Keeper snapshots. The interserver protocol in ClickHouse Keeper is also incompatible with ZooKeeper so a mixed ZooKeeper / ClickHouse Keeper cluster is impossible.
```

nit: This sentence was okay before, but if we we're going to change the punctuation I would use a colon here or an em dash **—**

SpencerTorres (changes requested):

SpencerTorres:
The table looks much cleaner, thanks for updating it 🚀

SpencerTorres (approved):

PR #2437, opened 2024-06-28
Add documentation for S3 continuous ingestion for clickpipes

tpanetti:
This adds a quick blurb about using S3 continuous ingestion and removes the sections saying we don't support it. Also updates the screenshot to show continuous ingestion slider enabled.


docs/en/integrations/data-ingestion/clickpipes/object-storage.md
```diff
@@ -104,6 +104,9 @@ This is not a valid path. `{N..M}` are not supported in ClickPipes.
 https://datasets-documentation.s3.eu-west-3.amazonaws.com/http/{documents-01,documents-02}.ndjson.gz
 :::
 
+## Continuous Ingest
+ClickPipes supports continuous ingestion from both S3 and GCS. When enabled, ClickPipes will continuously ingest data from the specified path. However, new files must be lexically greater than the last ingested file, meaning they must be named in a way that defines the ingestion order. For instance, files named `file1`, `file2`, `file3`, etc., will be ingested sequentially. If a new file is added with a name like `file0`, ClickPipes will not ingest it because it is not lexically greater than the last ingested file.
```
pjhampton, line 108: This reads well. I like the example. 

 1. Do we support the `_path` and `_file` virtual columns?
 2. Should we document possible scenarios of duplicated rows?

Maybe that is for the s3 pipe in general, than just streaming import

pjhampton (approved):

PR #2452, opened 2024-07-02
Add limitation on x-cloud role auth for S3 clickpipes.

pjhampton:


docs/en/integrations/data-ingestion/clickpipes/object-storage.md
```diff
@@ -87,6 +87,7 @@ The supported formats are:
 
 ## Limitations
 
+- Role authentication is not available for S3 ClickPipes for ClickHouse instances deployed into GCP or Azure. It is only supported for AWS CH Cloud instances.
```
kitop, line 90: We don't seem to use the `CH` shorthand anywhere in docs, so a small tweak:
```suggestion
- Role authentication is not available for S3 ClickPipes for ClickHouse instances deployed into GCP or Azure. It is only supported for AWS ClickHouse Cloud instances.
```

PR #2470, opened 2024-07-07
Add supported Spark <> ClickHouse types

BentsiLeviav:


mzitnik (changes requested):
Can we aline the layout like https://clickhouse.com/docs/en/integrations/kafka/clickhouse-kafka-connect-sink#supported-data-types

mzitnik (approved):

PR #2485, opened 2024-07-12
Add note about S3 virtual columns for clickpipes and integrations

kitop:
Virtual columns are supported in ClickPipes, so added a callout about this with a link to the list of virtual columns.
S3 Integration already mentioned it, but this adds a link to the page so it's more discoverable what virtual columns are available.

### DIFF
--- a/docs/en/integrations/data-ingestion/clickpipes/object-storage.md
+++ b/docs/en/integrations/data-ingestion/clickpipes/object-storage.md
@@ -44,6 +44,10 @@ You have familiarized yourself with the [ClickPipes intro](./index.md).
 
   ![Use and existing table](./images/cp_step4b.png)
 
+:::info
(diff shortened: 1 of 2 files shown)


genzgd (approved):

docs/en/integrations/data-ingestion/clickpipes/object-storage.md
```diff
@@ -44,6 +44,10 @@ You have familiarized yourself with the [ClickPipes intro](./index.md).
 
   ![Use and existing table](./images/cp_step4b.png)
 
+:::info
+You can also map [virtual columns](../../sql-reference/table-functions/s3#virtual-columns), like `_path` or `_size`, to fields.
```
mshustov, line 48: optional: we could unify the representation. see Kafka docs https://github.com/ClickHouse/clickhouse-docs/pull/2260

PR #2527, opened 2024-07-25
TTL Example: Removing Rows Using Where Condition

allegrinisante:
Added an example demonstrating how to remove rows based on a field value.


SpencerTorres (approved):
Didn't know this, great example 💪

PR #2617, opened 2024-09-02
Fix SQL console role spelling

cwurm:
Same as https://github.com/ClickHouse/clickhouse-docs/pull/2576


SpencerTorres (approved):
Much better, thanks for updating this and #2576

PR #2621, opened 2024-09-04
Fixing Secure S3 documentation and IP filtering

santrancisco:
Fixing doc for the new changes with Secure S3 role name. It is no longer required to have `ClickHouseAccess-` prefix and make it easier for users to reuse existing roles in their environment.

Also fixed Ip filtering documentation as i came across it


docs/en/cloud/security/setting-ip-filters.md
```diff
@@ -69,31 +74,6 @@ curl https://<HOSTNAME>.clickhouse.cloud:8443
 Ok.
 ```
 
-## Importing and exporting filters
-From the **Security** tab you can also share (import or export) your filters.
```
santrancisco, line 73: Need sanity check here - @guykoh  I no longer see ability to export/import filter - is it hidden somewhere else or we removed it as part of unified console?

guykoh: Good question, I don't see it either, and vaguely remember we talked about it during the migration to the unified console. @serdec, @vineethasok, @ArctypeZach did we decide to remove the import/export of IP filtering? Or we just decided to add it later?

santrancisco: I see - then it's not just me missing feature flag or something, let's discuss it more later :) i will merge this doc update first to reflect the change in Secure S3 and ip filtering.

PR #2644, opened 2024-09-17
Update upgrades.md

santrancisco:
## Summary

Related to a slack community question in #help ... It seems we may cause some confusion here with the use of "production workload" and "production tier" in 2 bullet points next to each other


SpencerTorres (approved):
Just saw this question in the slack, I can see why users would get confused with these terms immediately next to one another

Instead of "non-production environment" we could say "non-critical workloads", or some other phrase that indicates the lack of stability in early/fast upgrades

PR #2694, opened 2024-10-04
Warn about Datadog Custom Metrics

tzjames:
## Summary
There are 1410 `TYPE Clickhouse` matches in my endpoint's output, which would make it 4230 custom metrics which could end up costing some $200 a month or more.  It seems wise to alert people to this, so they can decide for themselves whether it is worth it.

### DIFF
--- a/docs/en/integrations/prometheus.md
+++ b/docs/en/integrations/prometheus.md
@@ -240,7 +240,7 @@ Note the `honor_labels` configuration parameter needs to be set to `true` for th
 
 ## Integrating with Datadog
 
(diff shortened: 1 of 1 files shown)


pjhampton:
Thanks for the contribution, @tzjames! Can you please sign the CLA?

tzjames:
@pjhampton Thanks done.  Ideally there would be a ClickhouseCloud Datadog integration so that nothing is counted as a custom metric.  Since there is the Prometheus endpoint, I would think it should be fairly straightforward to make the integration.  Are there any plans to do so?

tbragin:
cc @krithika-ch

justindeguzman (approved):

PR #2697, opened 2024-10-05
fix syntax error in python client docs

wstswsb:
## Summary
Remove leading zero when creating datetime, which caused SyntaxError


genzgd:
Thanks!

PR #2704, opened 2024-10-15
Remove ANSI from sidebar

alexey-milovidov:
## Summary

It is confusing, and I was not able to find what is `E111 | Single row SELECT statement`.
This is from the official standard document that is difficult to find and use.

## Checklist


alexey-milovidov:
Ha-ha, this is:

```
SELECT a, b, c INTO @var FROM table
```

No way we need this. Good riddance.

https://documentation.softwareag.com/adabas/cxx1103/embedded/SELECT_(SINGLE_ROW).htm

PR #2731, opened 2024-10-25
note on max_bytes_to_merge_at_max_space_in_pool

gingerwizard:
## Summary

Note on RMT limitations


SpencerTorres (approved):
Good info. Very similar to this page: https://clickhouse.com/docs/en/cloud/bestpractices/avoid-optimize-final

In a future update perhaps we could consolidate this info and expand on it in the `OPTIMIZE` docs page. Seems like `FINAL` may need its own section: 
https://clickhouse.com/docs/en/sql-reference/statements/optimize

PR #2800, opened 2024-11-18
add Japanese localization scripts

matsumotomiki:
## Summary
add an initial script to localize docs/en to docs/ja


gingerwizard:
Going to merge this, @matsumotomiki given these scripts don't impact the current build. Ideally we'd encapsulate the entire translation process into a single "generate_japenese_docs" script which handled the manual steps you document.

PR #2823, opened 2024-11-24
Fix ugly looking navigation on mobile

Blargian:
## Summary
I'm sure there are very few people looking at ClickHouse docs on mobile... never the less, the secondary (docs) navigation bar looks horrible on mobile:

image

This PR adds custom React components `Hamburger`, `HamburgerMenu` and `MobileMenu` (grouped under `DocsCategoryMobileNav`) such that the secondary menu acts in a similar way to the primary menu (displays a burger menu and navigation overlay for smaller screen sizes).

**Result**

demo

Still to do:

- [x] Fix search bar styling for dark theme.
- [x] Hide navbar when link is clicked.
- [x] Move `dropdownCategories` in `src/theme/NavBar/Content/index.js` to it's own config file, accessible by this new component.

## Checklist

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -1399,6 +1399,286 @@ const sidebars = {
       dirName: "zh",
     },
   ],
-};
+  // Used for generating the secondary nav mobile menu (DocsCategoryDropdown)
+  dropdownCategories : [
+    {
+      type: 'category',
+      label: 'Getting Started',
+      //description: 'Learn how to use ClickHouse',
+      items: [
+        {
+          type: 'link',
+          label: 'Introduction',
+          //description: 'An introduction to ClickHouse',
+          href: '/docs'
+        },
+        {
+          type: 'link',
+          label: 'Starter Guides',
+          //description: 'Start here when learning ClickHouse',
+          href: '/docs/en/guides/creating-tables'
+        },
+        {
+          type: 'link',
+          label: 'Concepts',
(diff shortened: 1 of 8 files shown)


justindeguzman (approved):

justindeguzman:
Thanks @Blargian. I'm going to merge this in as an improvement, but my suggest would be to:

- Flip the menus so that the top-level categories are the first menu, and the second menu opens the navbar for the current category
- For the top-level menu, it maybe should not be an icon but instead "Getting Started", "Cloud", etc. The double menus may look confusing

Screenshot 2024-11-25 at 11 44 07 AM

Blargian:
> Thanks @Blargian. I'm going to merge this in as an improvement, but my suggest would be to:
> 
> * Flip the menus so that the top-level categories are the first menu, and the second menu opens the navbar for the current category
> * For the top-level menu, it maybe should not be an icon but instead "Getting Started", "Cloud", etc. The double menus may look confusing
> 
> Screenshot 2024-11-25 at 11 44 07 AM

Thanks @justindeguzman, that makes sense. I'm thinking now maybe just a change to the styling of the menu you implemented for mobile is required: 
image

The secondary menu on mobile could show the categories: `Getting started`, `Cloud`, `Managing Data` etc and instead of the sub-menu popping up on hover, it could slide out from the left of the screen or something.  

I'll see what can be done.

Issue #2861, opened 2024-12-06
Lightmode - text boxes don't show


gingerwizard:
reported by @linhgiang24 

Cloud security features dont show in light mode.

https://clickhouse.com/docs/en/cloud/security/shared-responsibility-model#clickhouse-cloud-configurable-security-features

![image](https://github.com/user-attachments/assets/d7583cb2-3602-4e5b-bda6-f87c68299a74)

gingerwizard:
https://github.com/ClickHouse/clickhouse-docs/pull/2864

PR #2864, opened 2024-12-09
fix rendering for light mode on details table

gingerwizard:
## Summary

Fixes https://github.com/ClickHouse/clickhouse-docs/issues/2861

## Before

image

image

## After

image
image

## Checklist


gjones:
This does fix the issue, so I guess we can merge. I'm unsure why the accordions are black in light mode anyway though. Ideally we'd just use Click UI accordions here. Maybe we can merge this and make a separate issue around transitioning the docs to using Click UI accordions.

gjones (approved):

PR #2893, opened 2024-12-12
fast release is now self-serve for production tier

allegrinisante:
"Fast Release" is only available for the production tier and is a self-serve process accessible through the UI. Rewording this as the previous message may cause ambiguity for customers using the development tier, where the UI does not support the Fast Release option.

### DIFF
--- a/docs/en/cloud/manage/upgrades.md
+++ b/docs/en/cloud/manage/upgrades.md
@@ -23,7 +23,7 @@ You will not be charged for the time that the service is under maintenance. _Mai
 
 ## Release channels (upgrade schedule)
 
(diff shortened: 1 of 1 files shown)


docs/en/cloud/manage/upgrades.md
thomoco, line 1: Fast Release channel is allowed for Production services. We need to review this wording:

```
You are able to specify the upgrade schedule for your ClickHouse Cloud service by subscribing to a specific release channel. Besides the regular upgrade schedule, we offer a **Fast release** channel if you would like your services to receive updates ahead of the regular release schedule. Please be aware that the **Fast release** channel can encounter previously undiscovered issues or bugs, although can be used for production or critical services for those users looking to use the latest features.
```

PR #2903, opened 2024-12-14
[AWS] Update for "Multiple PrivateLinks in one region"

tsolodov:
## Summary

subj

### DIFF
--- a/docs/en/cloud/security/aws-privatelink.md
+++ b/docs/en/cloud/security/aws-privatelink.md
@@ -374,7 +374,7 @@ In this example connection to `xxxxxxx.yy-xxxx-N.vpce.aws.clickhouse.cloud` host
 
 ### Multiple PrivateLinks in one region
 
-If you require two or more AWS Private Links within the same AWS region, then please note: In ClickHouse, we have a VPC Endpoint service at a regional level. When you setup two or more VPC Endpoints in the same VPC - from the AWS VPC perspective - you are utilizing just a single AWS Private Link. In such a situation where you need two or more AWS Private Links configured within the same region, please just create just one VPC Endpoint in your VPC, and request that ClickHouse configure the same VPC Endpoint ID for all of your ClickHouse services in the same AWS region.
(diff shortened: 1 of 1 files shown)


docs/en/cloud/security/aws-privatelink.md
```diff
@@ -374,7 +374,7 @@ In this example connection to `xxxxxxx.yy-xxxx-N.vpce.aws.clickhouse.cloud` host
 
 ### Multiple PrivateLinks in one region
 
-If you require two or more AWS Private Links within the same AWS region, then please note: In ClickHouse, we have a VPC Endpoint service at a regional level. When you setup two or more VPC Endpoints in the same VPC - from the AWS VPC perspective - you are utilizing just a single AWS Private Link. In such a situation where you need two or more AWS Private Links configured within the same region, please just create just one VPC Endpoint in your VPC, and request that ClickHouse configure the same VPC Endpoint ID for all of your ClickHouse services in the same AWS region.
+In most cases, you only need to create a single endpoint service per VPC. The endpoint connection can route requests from the VPC to multiple ClickHouse cloud services. Your ClickHouse Cloud services might be sharded within a region across multiple environments. Each environment uses a single endpoint configuration. In this case, you may have more than one endpoint per VPC. Each endpoint routes requests to the appropriate environment, while DNS configuration directs requests to the specific endpoint.
```
gingerwizard, line 377: "services might be sharded within a region"

this is a confusing use of the word "sharded" @tsolodov maybe split is better?

also i think we want to use the word However i.e.

In most cases, you only need to create a single endpoint service for each VPC. This endpoint can route requests from the VPC to multiple ClickHouse Cloud services. However, if your ClickHouse Cloud services are sharded within a region and span multiple environments, each environment will require its own endpoint configuration. In such cases, you may need multiple endpoints per VPC. Each endpoint routes requests to its respective environment, while DNS configuration ensures requests are directed to the correct endpoint.

gingerwizard (changes requested):

docs/en/cloud/security/aws-privatelink.md
tsolodov: let's keep is simple for now:

`In most cases, you only need to create a single endpoint service for each VPC. This endpoint can route requests from the VPC to multiple ClickHouse Cloud services.`

gingerwizard (approved):

PR #2922, opened 2024-12-17
Update index.md

aashishkohli:
Updated FAQs to include self-serve org switching process from the cloud  UI

## Checklist

### DIFF
--- a/docs/en/cloud/manage/billing/marketplace/index.md
+++ b/docs/en/cloud/manage/billing/marketplace/index.md
@@ -34,7 +34,11 @@ You can switch between organizations from the bottom left menu of the ClickHouse
 
 ### I am an existing ClickHouse Cloud user. What should I do if I want my existing services to be billed via marketplace?​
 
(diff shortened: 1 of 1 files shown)


ramseyCH:
`NOTE: Your existing services and organizations from any prior ClickHouse Cloud signups will remain and they will not be connected to the marketplace billing.`

This line could be updated also.

ramseyCH:
Sorry my comment could have been clearer. I was quoting the line from the FAQ question above the one you updated. I think this line is now obsolete.. since the new MP onboarding process gives them the option of associating their existing services to the marketplace. So we can probably just delete the line I quoted

aashishkohli:
I thought you were commenting that only the org that is being linked will bill through the MP and not the other orgs they have. Was that not the intent of the original comment? I can certainly take it out if it sounds confusing.

ramseyCH:
I re-read it and it looks good to me.

ramseyCH (approved):

PR #2930, opened 2024-12-19
Parts remove reference

tom-clickhouse:
Remove parts reference from tutorial, as that concept is not required there.


gingerwizard:
Dont merge. @tom-clickhouse can you make a PR with just the changed file please.

PR #2941, opened 2024-12-20
Backfill guide

gingerwizard:
## Summary

Guide on how to backfill data inc. mvs. cc @tom-clickhouse 

## Checklist


docs/en/data-modeling/backfilling.md
```diff
@@ -0,0 +1,627 @@
...
+```
+
+While we can add the target table, prior to adding the materialized view we modify its `SELECT` clause to include a filer which only considers rows greater than an arbitary time in the near future - in this case we assume `2024-12-17 09:00:00` is a few minutes in the future.
+
+```sql
+CREATE MATERIALIZED VIEW pypi_downloads_per_day_mv TO pypi_downloads_per_day
+AS SELECT
+ toStartOfHour(timestamp) as hour,
+ project, count() AS count
+FROM pypi WHERE timestamp < '2024-12-17 09:00:00'
```
tom-clickhouse, line 373: Filter needs to use > instead of <
WHERE timestamp > '2024-12-17 09:00:00'

tom-clickhouse: actually >=

tom-clickhouse (commented):
slight logic issue

Issue #2940, opened 2024-12-20
Improve readability of documentation


Blargian:
ClickHouse documentation has a rather large character count per line. With browser zoom set to 100%, taking [this page](https://clickhouse.com/docs/en/updating-data) as an example, the character count with spaces is 219. A count somewhere between 50 - 75 is more optimal for readability ([source](https://baymard.com/blog/line-length-readability)). 

Compare for example ClickHouse docs to ReactNative, MongoDB, Jest docs:

![image](https://github.com/user-attachments/assets/cd44f8bb-e428-4264-ba4a-b9999a1f24da)
(character count with spaces: 219)

![image](https://github.com/user-attachments/assets/bb229c6b-7a2f-491c-a559-9f61bcf2bb5a)
(character count with spaces: 99)

![image](https://github.com/user-attachments/assets/082f0735-b80d-4e87-91a7-8cc41d3eff95)
(character counts with spaces: 94)

![image](https://github.com/user-attachments/assets/f58813d4-cb74-4de9-9f91-cf57f515b795)
(character count: 148)

Pages will be more readable with a narrower content area.

Blargian:
Closing, this is a more general issue that design is aware of and will address.

PR #2944, opened 2024-12-22
Make content more readable

Blargian:
## Summary

Introduces a small styling change to content width to improve readability. Closes [#2940](https://github.com/ClickHouse/clickhouse-docs/issues/2940). Currently docs have the Docusaurus content area defaults overridden to make content full page width. This results in a high character count per line compared to other documentation sites (see eg. React Native, Jest, MongoDB) and general guidelines for readability (see #2940 for justification).

Before: 
image

After: 
image

Small change but makes a huge difference to readability in my opinion. We can always increase padding between content and the table of contents on the right if the white space on the right feels uncomfortable.

Of course subjective, interested to know what others think.

## Checklist

### DIFF
--- a/src/css/custom.scss
+++ b/src/css/custom.scss
@@ -419,6 +419,12 @@ li.theme-doc-sidebar-item-category-level-2 .menu__list-item-collapsible {
   background: #f8f8f8;
 }
 
+/* override content width */
+:root {
+  --ifm-container-width: 1280px;
+  --ifm-container-width-xl: 1600px;
+}
+
 /* the h1 style on the category pages overrides the font-size */
 .title_node_modules-\@docusaurus-theme-classic-lib-next-theme-DocCategoryGeneratedIndexPage-styles-module {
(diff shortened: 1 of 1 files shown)


gingerwizard:
@gjones your opinion here please

gingerwizard:
@Blargian we defer this to design. cc @gjones who has to approve visual changes.

Blargian:
Actually, docs site looks quite okay on a smaller screen. I think all that's needed here is a media query for devices with wider screens (eg 2560x1600 was what I was using when I made the issue)

Blargian:
> @Blargian Thanks for doing this! I'd rather we avoided centring the content though as it introduces an awkward gap between the navigation and the documentation. I do agree that the width needs to be limited though. I'd vote we keep it simple and just a `max-width: 1024px` (or larger) to the `docItemContainer_c0TR` which would give us something like below. Would that work for you?
> 
> CleanShot 2025-01-06 at 13 25 24@2x

That works too!

PR #2969, opened 2025-01-01
Add `CollapsingMergeTree` to updates page

Blargian:
## Summary
The blog post linked on the [Updating Data](https://clickhouse.com/docs/en/updating-data) page mentions use of `CollapsingMergeTree` as an approach to updates. We however do not list it as an approach on this page.

- Added `CollapsingMergeTree` to the table
- Added a section with short explanation and simple example from the docs for `CollapsingMergeTree`
- Renamed the light weight deletes section and added the "available in cloud" badge.

Documentation page for `CollapsingMergeTree` improved via [74062](https://github.com/ClickHouse/ClickHouse/pull/74062)

## Checklist


docs/en/managing-data/updates.md
```diff
@@ -76,6 +81,51 @@ Note that for lightweight updates, a mutation is still used to update the data;
 
 Read more about [lightweight updates](/en/guides/developer/lightweight-update).
 
+## Collapsing Merge Tree
+
+Stemming from the idea that updates are expensive but inserts can be leveraged to perform updates,
```
gingerwizard, line 86: is there a reason you use line breaks?

docs/en/managing-data/updates.md
```diff
@@ -76,6 +81,51 @@ Note that for lightweight updates, a mutation is still used to update the data;
...
+FROM UAct
+GROUP BY UserID
+HAVING sum(Sign) > 0
+
+┌──────────────UserID─┬─PageViews─┬─Duration─┐
+│ 4324182021466249494 │         6 │      185 │
+└─────────────────────┴───────────┴──────────┘
+```
+
+:::note
```
gingerwizard, line 122: They key thing here is it requires the user to maintain state client side - while its the most efficient for clickhouse, this complex at scale. We should state this.

gingerwizard (changes requested):
@Blargian thanks, minor changes and clarification. why line returns btw?

docs/en/managing-data/updates.md
Blargian: @gingerwizard are you referring to the edits to the query results here?

Blargian: ah, you're probably referring to the line breaks in the text itself? If so it comes from the CLion plugin I am using, it suggests where to place a line break such that the markdown is readable. space + /n will push the text to a new line, but /n without a space before it will still render on the same line, but looks more readable in the document itself when editing.

docs/en/managing-data/updates.md
Blargian: Update pushed

PR #2973, opened 2025-01-04
Replace sed with cat

cwurm:
The `sed` command doesn't do what it's probably supposed to do. To match lines from the beginning of the file until the pattern it needs to start with `1,/` not `0,/`. But also, I don't think we want this since it would remove the table of contents which is useful to have. This is probably some artefact from before something changed in the formatting of the changelog. 

As it is, it just appends the contents, which we can do easily with `cat`.


gingerwizard:
Fixed already

Issue #3038, opened 2025-01-07
Upgrade to Docusaurus v3


gingerwizard:
This is needed not least to address security issues in dependencies.

Number of tasks:

- [ ] Use .mdx extension whenever we use JSX, import, or export. In future versions of Docusaurus, .md files will be parsed as standard [CommonMark](https://commonmark.org/).
- [ ] MDX3 is stricter, address compilation issues - common ones are [here](https://docusaurus.io/blog/preparing-your-site-for-docusaurus-v3#common-mdx-problems). Run `npx docusaurus-mdx-checker`
- [ ] Upgrade all packages
- [ ] Upgrade node to v18/v20 LTS
- [ ] Upgrade to React v18.0+, MDX v3.0+, TypeScript v5.1+, prism-react-renderer v2.0+, react-live v4.0+, remark-emoji v4.0+, mermaid v10.4+
- [ ] Admonian colors have changed - [specifically warning](https://docusaurus.io/blog/preparing-your-site-for-docusaurus-v3#admonition-warning)

As a result of possible visual changes, we will implement visual testing first. We plan to use this moving forward for all package upgrades.

- [ ] Implement visual testing - this is needed for all future package dependencies.

gingerwizard:
Done!

PR #3044, opened 2025-01-07
visual testing

gingerwizard:
## Summary

Adds visual testing using playwright and argo. We have to upgrade some packages for docusaurus to make this work, moving to the latest v2 as well as node v20 LTS. From what i can tell this works - moving forward, we wont have this uncertainty!

https://github.com/ClickHouse/clickhouse-docs/issues/3038

## Checklist


gingerwizard:
We need this https://github.com/ClickHouse/ClickHouse/pull/74268

PR #3057, opened 2025-01-08
Use new search link APi

mwoenker:
## Summary

It seems like we are importing an API that is no longer exported from the docusaurus package. I believe I've found a new hook that serves the same purpose.

Fixes https://github.com/mwoenker/clickhouse-docs/pull/new/mw-fix-search-crash

## Checklist


src/theme/SearchBar/index.js
rfraposa, line 1: Hmm....I won't be much help in reviewing this. I don't remember how this go setup in the first place! 😄 But as long as the search still works, then it LGTM.

PR #3067, opened 2025-01-10
upgrade to docusaurus v3

gingerwizard:
## Summary

Upgrade to docusaurus v3. Updates packages, handles any build issues (mdx and package dependencies) and adds truncate markers for the KB. 

We remove the `remark-docusaurus-tabs` - its not supported in v2 and wasn't used. Other changes are mostly around need to use es modules in v3.

Note: i added a title and description to every KB while i was editing. Eventually we should use these and make it searchable.

Depends on https://github.com/ClickHouse/ClickHouse/pull/74423 so draft until this is merged.

Note ive turned off the link checker as it fails heavily for ru and zh docs. These docs are likely to be regenerated and replaced so we don't want address these. Once regenerated we'll turn on link checker to throw.

## Checklist


docs/en/cloud/manage/api/invitations-api-reference.md
```diff
@@ -5,5 +5,148 @@ title: Invitations
 
 ## List all invitations
 
-This file is generated by `clickhouseapi.js` during the build process.  If the 
-content needs changing please edit `clickhouseapi.js`.
+Returns list of all organization invitations.
```
Blargian, line 8: These and other API files shouldn't get checked in

Blargian (approved):
Apart from the checked in API files LGTM

PR #3071, opened 2025-01-13
Update JDBC docs for V2

Paultagoras:
## Summary
Added docs to support the V2 version (while keeping the V1 version of the docs)

Closes https://github.com/ClickHouse/clickhouse-java/issues/1480
## Checklist

### DIFF
--- a/docs/en/integrations/language-clients/java/jdbc-v2.md
+++ b/docs/en/integrations/language-clients/java/jdbc-v2.md
@@ -13,8 +13,8 @@ import CodeBlock from '@theme/CodeBlock';
 
 # JDBC Driver
 
-`clickhouse-jdbc` implements the standard JDBC interface using [client-v2](/docs/en/integrations/language-clients/java/client-v2.md).
-We recommend using [client-v2](/docs/en/integrations/language-clients/java/client-v2.md) directly if performance/direct access is critical.
+`clickhouse-jdbc` implements the standard JDBC interface using the latest [java client](/docs/en/integrations/language-clients/java/client-v2.md).
+We recommend using the latest [java client](/docs/en/integrations/language-clients/java/client-v2.md) directly if performance/direct access is critical.
 
 :::note
 If you're looking for a prior version of the JDBC driver docs, please see [here](/docs/en/integrations/language-clients/java/jdbc-v1.md).
@@ -23,13 +23,13 @@ If you're looking for a prior version of the JDBC driver docs, please see [here]
 ## Changes from 0.7.x
 In 0.8 we tried to make the driver more strictly follow the JDBC specification, so there are some removed features that may affect you:
 
-| Old Feature                      | Notes                                                                                                                                                                                                                                                                                                     |
-|----------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| Transaction Support              | Early versions of the driver only **simulated** transaction support, which could have unexpected results.                                                                                                                                                                                                 |
-| Response Column Renaming         | `ResultSet` was mutable - for efficiency sake they're now read-only                                                                                                                                                                                                                                       |
-| Multi-Statement SQL              | Multi-statement support was only **simulated**, now it strictly follows 1:1                                                                                                                                                                                                                               |
-| Named Parameters                 | Not part of the JDBC spec                                                                                                                                                                                                                                                                                 |
-| Stream-based `PreparedStatement` | Early version of the driver allowed for non-jdbc usage of `PreparedStatement` - if you desire such options, we recommend looking at [Client-V2](/docs/en/integrations/language-clients/java/client-v2.md) and its [examples](https://github.com/ClickHouse/clickhouse-java/tree/main/examples/client-v2). |
+| Old Feature                      | Notes                                                                                                                                                                                                                                                                                                           |
+|----------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Transaction Support              | Early versions of the driver only **simulated** transaction support, which could have unexpected results.                                                                                                                                                                                                       |
+| Response Column Renaming         | `ResultSet` was mutable - for efficiency sake they're now read-only                                                                                                                                                                                                                                             |
+| Multi-Statement SQL              | Multi-statement support was only **simulated**, now it strictly follows 1:1                                                                                                                                                                                                                                     |
+| Named Parameters                 | Not part of the JDBC spec                                                                                                                                                                                                                                                                                       |
+| Stream-based `PreparedStatement` | Early version of the driver allowed for non-jdbc usage of `PreparedStatement` - if you desire such options, we recommend looking at the [Java Client](/docs/en/integrations/language-clients/java/client-v2.md) and its [examples](https://github.com/ClickHouse/clickhouse-java/tree/main/examples/client-v2). |
 
 :::note
 `Date` is stored without timezone, while `DateTime` is stored with timezone. This can lead to unexpected results if you're not careful.
@@ -83,19 +83,19 @@ implementation 'com.clickhouse:clickhouse-jdbc:0.8.0:shaded-all'
 
 **Connection Properties**:
 
-Beyond standard JDBC properties, the driver supports the ClickHouse-specific properties offered by the underlying [client](/docs/en/integrations/language-clients/java/client-v2.md).
+Beyond standard JDBC properties, the driver supports the ClickHouse-specific properties offered by the underlying [java client](/docs/en/integrations/language-clients/java/client-v2.md).
 Where possible methods will return an `SQLFeatureNotSupportedException` if the feature is not supported. Other custom properties include:
 
 | Property                         | Default | Description                                                    |
 |----------------------------------|---------|----------------------------------------------------------------|
 | `disable_frameworks_detection`   | `true`  | Disable frameworks detection for User-Agent                    |
 | `jdbc_ignore_unsupported_values` | `false` | Suppresses `SQLFeatureNotSupportedException`                   |
-| `clickhouse.jdbc.v1`             | `false` | Use JDBC-V1 instead of JDBC-V2                                 |
+| `clickhouse.jdbc.v1`             | `false` | Use older JDBC implementation instead of new JDBC              |
 | `default_query_settings`         | `null`  | Allows passing of default query settings with query operations |
 
 ## Supported data types
 
-JDBC Driver supports the same data formats as the underlying [client](/docs/en/integrations/language-clients/java/client-v2.md).
+JDBC Driver supports the same data formats as the underlying [java client](/docs/en/integrations/language-clients/java/client-v2.md).
 
 ### Handling Dates, Times, and Timezones
 `java.sql.Date`, `java.sql.Time`, and `java.sql.Timestamp` can complicate how Timezones are calculated - though they're of course supported,
@@ -184,7 +184,7 @@ try (HikariDataSource ds = new HikariDataSource(poolConfig);
 ```
 
 ## More Information
-For more information, see our [GitHub repository](https://github.com/ClickHouse/clickhouse-java) and [Client-V2 documentation](/docs/en/integrations/language-clients/java/client-v2.md).
+For more information, see our [GitHub repository](https://github.com/ClickHouse/clickhouse-java) and [Java Client documentation](/docs/en/integrations/language-clients/java/client-v2.md).
 
 
 ## Troubleshooting


mshustov (changes requested):
1. It's discussable. Personally, I don't think we need to create another separate entry in the docs, as JDBC v2 is an internal name. For the end user it's the same driver with some behavioural changes from version `0.8.0`.
2. I didn't find a list of changes and unsupported features in JDBC starting from `0.8.0` (Transactions, for example), which is important to highlight. It's not necessarily to follow the same format, but you can see how it was done in the same JDBC driver  previously https://github.com/ClickHouse/clickhouse-java/tree/v0.4.0/clickhouse-jdbc#upgrade-to-032

laeg:
> 1. It's discussable. Personally, I don't think we need to create another separate entry in the docs, as JDBC v2 is an internal name. For the end user it's the same driver with some behavioural changes from version `0.8.0`.

The tricky part here is the difference between a SaaS product (always showing latest rather than maintaining state) and traditional SW documentation.

But I agree with this sentiment, it also might be worth us exploring at this point how do we do "versioning" of the documentation rather than having "extra pages" - like having a version drop down and corresponding version in the URL.

Maybe we need to consider something like having JavaDocs being the "source of truth" of functionality?

At least one of the immediate things we should do is treat the sidebar as the "ideal customer journey" rather than an index. We can link in the main doc to the v1 documentation until it's deprecated.

> 2. I didn't find a list of changes and unsupported features in JDBC starting from `0.8.0` (Transactions, for example), which is important to highlight. It's not necessarily to follow the same format, but you can see how it was done in the same JDBC driver  previously https://github.com/ClickHouse/clickhouse-java/tree/v0.4.0/clickhouse-jdbc#upgrade-to-032

I think rather than listing out all unsupported features, it would be more beneficial for a user to have a "migration guide" that has this information in and alternatives to achieving similar functionality if applicable.

chernser:
@mshustov I think separating documents is good because:
- less confusion because information is separated by version and it is clear what should work in v2, v1
- when jdbc v1 is deprecated - we can remove v1 doc easily and revert as needed
- visit statistics.

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,168 @@
...
+import TabItem from '@theme/TabItem';
+import CodeBlock from '@theme/CodeBlock';
+
+
+# JDBC Driver
+
+`clickhouse-jdbc` implements the standard JDBC interface. Being built on top of [clickhouse-client](/docs/en/integrations/sql-clients/sql-console), it provides additional features like custom type mapping, transaction support, and standard synchronous `UPDATE` and `DELETE` statements, etc., so that it can be easily used with legacy applications and tools.
+
+:::note
+    Latest JDBC (0.8.0) version uses Client-V2 by default.
```
chernser, line 19: I'd say that JDBC v2 uses only client v2. 
because  the statement is not completely true for whole JDBC 0.8.0

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,168 @@
...
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import CodeBlock from '@theme/CodeBlock';
+
+
+# JDBC Driver
+
+`clickhouse-jdbc` implements the standard JDBC interface. Being built on top of [clickhouse-client](/docs/en/integrations/sql-clients/sql-console), it provides additional features like custom type mapping, transaction support, and standard synchronous `UPDATE` and `DELETE` statements, etc., so that it can be easily used with legacy applications and tools.
```
chernser, line 16: it is build on top of client-v2

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,168 @@
...
+```groovy
+// https://mvnrepository.com/artifact/com.clickhouse/clickhouse-jdbc
+// use uber jar with all dependencies included, change classifier to http for smaller jar
+implementation 'com.clickhouse:clickhouse-jdbc:0.8.0:shaded-all'
+```
+
+</TabItem>
+</Tabs>
+
+Since version `0.5.0`, we are using Apache HTTP Client that's packed the Client. Since there is not a shared version of the package, you need to add a logger as a dependency.
```
chernser, line 65: no need to add - this is already shade-all

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,168 @@
...
+**Connection Properties**:
+
+Beyond standard JDBC properties, the driver supports the ClickHouse-specific properties offered by the underlying [client](/docs/en/integrations/language-clients/java/client-v2.md).
+Where possible methods will return an SQLFeatureNotSupportedException if the feature is not supported. Other custom properties include:
+
+| Property                         | Default | Description                                 |
+|----------------------------------|---------|---------------------------------------------|
+| `disable_frameworks_detection`   | `true`  | Disable frameworks detection for User-Agent |
+| `jdbc_ignore_unsupported_values` | `false` | Suppresses SQLFeatureNotSupportedException  |
+| `clickhouse.jdbc.v1`             | `false` | Use JDBC-V1 instead of JDBC-V2              |
```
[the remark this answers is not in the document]
chernser: missing `default_query_settings`
please check with `com.clickhouse.jdbc.internal.DriverProperties`

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
...
+
+**URL Syntax**: `jdbc:(ch|clickhouse)[:<protocol>]://endpoint1[,endpoint2,...][/<database>][?param1=value1&param2=value2][#tag1,tag2,...]`, for example:
+
+- `jdbc:clickhouse:http://localhost:8123`
+- `jdbc:clickhouse:https://localhost:8443?ssl=true`
+
+**Connection Properties**:
+
+Beyond standard JDBC properties, the driver supports the ClickHouse-specific properties offered by the underlying [client](/docs/en/integrations/language-clients/java/client-v2.md).
+Where possible methods will return an SQLFeatureNotSupportedException if the feature is not supported. Other custom properties include:
```
mshustov, line 69:
```suggestion
Where possible methods will return an `SQLFeatureNotSupportedException` if the feature is not supported. Other custom properties include:
```

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
...
+| Property                         | Default | Description                                                    |
+|----------------------------------|---------|----------------------------------------------------------------|
+| `disable_frameworks_detection`   | `true`  | Disable frameworks detection for User-Agent                    |
+| `jdbc_ignore_unsupported_values` | `false` | Suppresses SQLFeatureNotSupportedException                     |
+| `clickhouse.jdbc.v1`             | `false` | Use JDBC-V1 instead of JDBC-V2                                 |
+| `default_query_settings`         | `null`  | Allows passing of default query settings with query operations |
+
+## Supported data types
+
+JDBC Driver supports same data formats as the client library does.
```
mshustov, line 80: Should there be a link to a list of supported types?

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
...
+
+**Connection Properties**:
+
+Beyond standard JDBC properties, the driver supports the ClickHouse-specific properties offered by the underlying [client](/docs/en/integrations/language-clients/java/client-v2.md).
+Where possible methods will return an SQLFeatureNotSupportedException if the feature is not supported. Other custom properties include:
+
+| Property                         | Default | Description                                                    |
+|----------------------------------|---------|----------------------------------------------------------------|
+| `disable_frameworks_detection`   | `true`  | Disable frameworks detection for User-Agent                    |
+| `jdbc_ignore_unsupported_values` | `false` | Suppresses SQLFeatureNotSupportedException                     |
```
mshustov, line 74:
```suggestion
| `jdbc_ignore_unsupported_values` | `false` | Suppresses `SQLFeatureNotSupportedException`                     |
```

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Supported data types
+
+JDBC Driver supports same data formats as the client library does. 
+
+:::note
+- AggregatedFunction - :warning: does not support `SELECT * FROM table ...`
+- Decimal - `SET output_format_decimal_trailing_zeros=1` in 21.9+ for consistency
+- Enum - can be treated as both string and integer
+- UInt64 - mapped to `long` (in client-v1)
```
mshustov, line 86: What does `in client-v1` mean? It might be confusing to introduce the fourth concept on the page (jdbc v1/v2, client v1/v2)

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
...
+    ps.executeBatch(); // stream everything on-hand into ClickHouse
+}
+```
+## Migrating from 0.7.x
+In general we tried to make the driver more strictly follow the JDBC specification, so there are some changes that may affect you:
+
+| Old Feature                     | Notes                                                                                                                                                                                                     |
+|---------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Transaction Support             | Early versions of the driver only **simulated** transaction support, which could have unexpected results.                                                                                                 |
+| Response Column Renaming        | ResultSets were read/write - for efficiency sake they're now read-only                                                                                                                                    |
```
mshustov, line 132:
```suggestion
| Response Column Renaming        | ResultSets were mutable - for efficiency sake they're now read-only                                                                                                                                    |
```

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
...
+| Multi-Statement SQL             | Statements would split multi-statements and execute, now it strictly follows 1:1                                                                                                                          |
```
mshustov, line 133: You can add that it used to be a simulation

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
...
+| Named Parameters                |                                                                                                                                                                                                           |
```
mshustov, line 134: Should we add that it's not supported as not the part of the JDBC spec?

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
...
+| Stream-based PreparedStatements | Early version of the driver allowed for non-jdbc usage of PreparedStatements - if you desire such options, we recommend looking to [Client-V2](/docs/en/integrations/language-clients/java/client-v2.md). |
```
mshustov, line 135: Could we add a link to a specific section or a code snippet?

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
...
+
+
+
+## More Information
```
mshustov, line 139: the following parts exist in JDBC v1, but not v2. Is that intentionally? 
https://clickhouse-docs-el5b4zr1v-clickhouse.vercel.app/docs/en/integrations/java/jdbc-v1#configuring-http-library

https://clickhouse-docs-el5b4zr1v-clickhouse.vercel.app/docs/en/integrations/java/jdbc-v1#connect-to-clickhouse-with-ssl

https://clickhouse-docs-el5b4zr1v-clickhouse.vercel.app/docs/en/integrations/java/jdbc-v1#resolving-jdbc-timeout-on-large-inserts

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
...
+try (Connection conn = dataSource.getConnection(...);
+    Statement stmt = conn.createStatement()) {
+    ResultSet rs = stmt.executeQuery("select * from numbers(50000)");
+    while(rs.next()) {
+        // ...
+    }
+}
+```
+
+## Insert
```
mshustov, line 115: Could we add a snippet with Hikari from https://github.com/ClickHouse/clickhouse-java/pull/2083 and close https://github.com/ClickHouse/clickhouse-java/issues/1480 ?

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
...
+| Response Column Renaming        | ResultSets were read/write - for efficiency sake they're now read-only                                                                                                                                    |
+| Multi-Statement SQL             | Statements would split multi-statements and execute, now it strictly follows 1:1                                                                                                                          |
+| Named Parameters                |                                                                                                                                                                                                           |
+| Stream-based PreparedStatements | Early version of the driver allowed for non-jdbc usage of PreparedStatements - if you desire such options, we recommend looking to [Client-V2](/docs/en/integrations/language-clients/java/client-v2.md). |
+
+
+
+## More Information
+For more information, see our [GitHub repository](https://github.com/ClickHouse/clickhouse-java) and [Client-V2 documentation](/docs/en/integrations/language-clients/java/client-v2.md).
+```
```
mshustov, line 141: I see a few issues with `docs` label. should we address them from the very beginning? https://github.com/ClickHouse/clickhouse-java/issues?q=is%3Aissue%20state%3Aopen%20label%3Adocs%20label%3Ajdbc-v2

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
...
+try (PreparedStatement ps = conn.prepareStatement("INSERT INTO mytable VALUES (?, ?)")) {
+    ps.setString(1, "test"); // id
+    ps.setObject(2, LocalDateTime.now()); // timestamp
+    ps.addBatch();
+    ...
+    ps.executeBatch(); // stream everything on-hand into ClickHouse
+}
+```
+## Migrating from 0.7.x
+In general we tried to make the driver more strictly follow the JDBC specification, so there are some changes that may affect you:
```
mshustov, line 127: There are notes, but no *actionable recommendations* for the migration are provided. for example:
Multi-Statement SQL  - break the multi-statement query and execute in the app one by one 
Named Parameters  - leverage positional arguments 
etc.

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
...
+| Response Column Renaming        | ResultSets were read/write - for efficiency sake they're now read-only                                                                                                                                    |
+| Multi-Statement SQL             | Statements would split multi-statements and execute, now it strictly follows 1:1                                                                                                                          |
+| Named Parameters                |                                                                                                                                                                                                           |
+| Stream-based PreparedStatements | Early version of the driver allowed for non-jdbc usage of PreparedStatements - if you desire such options, we recommend looking to [Client-V2](/docs/en/integrations/language-clients/java/client-v2.md). |
+
+
```
mshustov, line 137: Could we add a section with logging instructions?

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
...
+- AggregatedFunction - :warning: does not support `SELECT * FROM table ...`
+- Decimal - `SET output_format_decimal_trailing_zeros=1` in 21.9+ for consistency
+- Enum - can be treated as both string and integer
+- UInt64 - mapped to `long` (in client-v1) 
+:::
+
+## Creating Connection
+
+```java
+String url = "jdbc:ch://my-server:8123/system"; // use http protocol
```
mshustov, line 92: We can add a separate example with supplying clickhouse server and client settings

docs/en/integrations/language-clients/java/jdbc-v2.md
```diff
@@ -0,0 +1,141 @@
+---
+sidebar_label: JDBC V2
+sidebar_position: 4
+keywords: [clickhouse, java, jdbc, driver, integrate]
+description: ClickHouse JDBC driver V2
```
mshustov, line 5: > But I agree with this sentiment, it also might be worth us exploring at this point how do we do "versioning" of the documentation rather than having "extra pages" - like having a version drop down and corresponding version in the URL.

@laeg What about then having without these v1/v2 but using the client versioning? There will be 2 pages then
1. description: ClickHouse JDBC driver > v0.8.x (or ClickHouse JDBC driver latest)
2. description: ClickHouse JDBC driver v0.7.x

We can manually add more pages if necessary and they always transparently mapped to *JDBC release* version

docs/en/integrations/language-clients/java/jdbc-v2.md
mshustov: we can do it a in follow-up

mshustov (commented):
the last thing, I'd like to sort out is versioning in https://github.com/ClickHouse/clickhouse-docs/pull/3071/files#r1925045569

and (minor)moving data/time concerns to `Supported types` section

docs/en/integrations/language-clients/java/jdbc-v2.md
laeg: yeah i would prefer if we aligned it more to release versions rather than internal code versions. 

did we look into whether we could do something more elegant for versioning the docs?

laeg:
adding my comment again:

> At least one of the immediate things we should do is treat the sidebar as the "ideal customer journey" rather than an index. We can link in the "v2" doc to the "v1" documentation until it's deprecated.

we can do the same for java v1 & v2 documentation

docs/en/integrations/language-clients/java/jdbc-v2.md
mshustov: > did we look into whether we could do something more elegant for versioning the docs?

I didn't manage to find a solution in 30min.

mshustov:
> At least one of the immediate things we should do is treat the sidebar as the "ideal customer journey" rather than an index. We can link in the "v2" doc to the "v1" documentation until it's deprecated.

so you want something like this?

mshustov (approved):

PR #3090, opened 2025-01-15
Follow up PR for breadcrumbs

Blargian:
## Summary
- Links top level nav to landing pages.
- Adds description column to tables which were missing it.
- Removes the unneeded optimizations overview page (and adds a redirect for it)
- Orders the Cloud top level menu such that it matches the order of the items in the left side menu.

## Checklist


gingerwizard:
@Blargian see hash fixes :|

PR #3104, opened 2025-01-17
Add script to check for KB tags and add tags to all KB articles

Blargian:
## Summary

- Adds a script to check that KB articles are tagged.
- Adds tags to all KB articles.
- Modifies styling of Docusaurus Blog.

Still to do:
- Add a KB only search bar and ability to search by tags

## Checklist


src/css/custom.scss
```diff
@@ -1113,3 +1108,50 @@ nav[aria-label='Docs sidebar'] {
...
+  }
+}
+
+.blog-tags-list-page {
+
+  h1 {
+    margin-top: 25px;
+  }
+
+}
```
gingerwizard, line 1157: net: newline

scripts/knowledgebase_article_checker.py
```diff
@@ -0,0 +1,111 @@
...
+                        print(f"Error parsing YAML in '{filename}': {e}")
+                        incorrectly_tagged_files.append(filename)
+        except FileNotFoundError:
+            print(f"File '{filename}' not found.")
+            incorrectly_tagged_files.append(filename)
+    return {'correctly_tagged': correctly_tagged_files, 'incorrectly_tagged': incorrectly_tagged_files}
+
+def main():
+
+    allowed_tags = \
```
gingerwizard, line 83: lets move this to a file.

gingerwizard: simple list, one per line

gingerwizard (changes requested):
Can we call this in build and new-build please

Blargian:
> Can we call this in build and new-build please

Actually hadn't thought of this before, but it's tricker with python scripts. If I add it to new build and build the user will have to activate a virtual env and install the required libraries for the script. 

Do we really need it on build and new-build?

gingerwizard:
no call in check style only, its enough. Only downside is like spell check we have to run multiple cmds locally but thats k.

gingerwizard (approved):
LGTM

Issue #3088, opened 2025-01-15
Settings fix for mdx3 errors


gingerwizard:
Currently settings generation causes mdx3 violations which blocked docusaurus v3 update. To work around this we fix manually in generation script.

https://github.com/ClickHouse/clickhouse-docs/blob/62ad30b3b88232ed7d57a9b6c9d7e7643c65d138/scripts/settings/autogenerate-settings.sh#L81-L102

Ideally this wouldn't be needed. cc @alexey-milovidov we can fix these settings if you're ok enforcing these escape rules on the team.

cc @Blargian

PR #3112, opened 2025-01-21
Remove string replacement from autogenerate settings script

Blargian:
## Summary
Closes https://github.com/ClickHouse/clickhouse-docs/issues/3088

## Checklist


Blargian:
Changes have been made to source but Vercel build still fails. I'm guessing the problem is that we use the CH binary in the autogenerate script but it's not built using the latest changes yet.

```
# Install ClickHouse
if [ ! -f ./clickhouse ]; then
  echo -e "[$SCRIPT_NAME] Installing ClickHouse binary\n"
  curl https://clickhouse.com/ | sh
fi
```

PR #3133, opened 2025-01-23
Add Github action for autogenerating table of contents json files

Blargian:
## Summary

Adds a Github action for generating table of contents .json files. These will be used to automatically generate table of contents tables on the landing pages.

Updates the current script to be able to:
- accept arg `--dir`
- generate one table of contents file for all files in a folder and it's subfolders (eg - as we would need for `interfaces/formats` with arg `--single-toc`)
- generate one table of contents per folder (the usual use case, default)

Future improvements:

- [x] Extend the script to be able to accept a single file which will list directories to generate for and whether to do a single toc or one per subdirectory. eg:

```text
/en/cloud/manage
/en/interfaces/formats --single-toc
```

Our folder structure's aren't great so for the time being we will add folders one by one.

## Checklist


Blargian:
Notes for self:

So, it works - `toc.json` is generated. This will work for docs on the ClickHouse/clickhouse-docs repohHowever the problem is that we git ignore any files from the ClickHouse/ClickHouse repo. So the question now is do we make these an exception and commit them? This is problematic I guess as we copy content across during build - they would get overwritten I assume.

Alternative solution: copy all table_of_contents files to their own folder maybe named as `interfaces_formats_toc.json` for example rather than just one `toc.json` in each folder.

.github/workflows/table_of_contents.yml
```diff
@@ -0,0 +1,79 @@
...
+  # Force the stdout and stderr streams to be unbuffered
+  PYTHONUNBUFFERED: 1
+
+on:
+  pull_request:
+    types:
+      - synchronize
+      - reopened
+      - opened
+
```
juliojimenez, line 17:
```suggestion
permissions:
  contents: write
```

.github/workflows/table_of_contents.yml
```diff
@@ -0,0 +1,75 @@
...
+            echo "has_changes=false" >> $GITHUB_OUTPUT
+          fi
+
+      # Step 7 - Commit and Push generated Table Of Contents files
+      - name: Commit and Push Changes
+        if: steps.check_changes.outputs.has_changes == 'true'
+        run: |
+          git add table-of-contents-files/*
+          git commit -m "Autogenerate table of contents files from GitHub action - $(date '+%Y-%m-%d %H:%M:%S')"
+          git push origin HEAD:update_table_of_contents
```
juliojimenez, line 75:
```suggestion
          git push origin HEAD:update_table_of_contents
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/table_of_contents.yml
```diff
@@ -0,0 +1,76 @@
...
+jobs:
+  generate_toc_formats:
+    runs-on: ubuntu-latest
+    steps:
+      # Step 1: Check out the repository
+      - name: Check out repository
+        uses: actions/checkout@v3
+        with:
+          repository: ClickHouse/clickhouse-docs
+          ref: ${{ github.ref }} # checkout the current branch
```
juliojimenez, line 28: These are defaults.

```suggestion
```

.github/workflows/table_of_contents.yml
```diff
@@ -0,0 +1,76 @@
...
+
+on:
+  pull_request:
+    types:
+      - synchronize
+      - reopened
+      - opened
+permissions:
+  contents: write
+jobs:
```
juliojimenez, line 19: let's try this to rule out permission issues

```suggestion
      - opened

permissions: write-all

jobs:
```

juliojimenez (approved):

PR #3168, opened 2025-01-29
Improve build-scripts

Blargian:
## Summary

- replaces `prep-from-local`, 'prep-from-master' and `copyClickhouseRepoDocs.sh` with a single script `scripts/copy-clickhouse-repo-docs.sh` which performs old `prep-from-master` by default or can perform `prep-from-local` if you provide it arg `-l`.
- `copy-clickhouse-repo-docs.sh` additionally copies files needed for any kind of autogeneration (changelog and settings generation) to `settings/tmp`. New files can be added to package.json "autogen_needed_files" to extend.
- `copy-clickhouse-repo-docs.sh` uses rsync, falls back to cp if it cannot find rsync
- Moves logic for generating changelog to own script.

To do:
- get rid of "new-build" and have just build
- add linter for python and bash from the main repo

## Checklist


scripts/settings/autogenerate-settings.sh
```diff
@@ -73,15 +83,14 @@ import BetaBadge from \'@theme/badges/BetaBadge\';
...
-if [ -f ./clickhouse ]; then
-  echo -e "\n[$SCRIPT_NAME] Deleting ClickHouse binary"
-  rm ./clickhouse
-fi
+echo "[$SCRIPT_NAME] Auto-generation of settings markdown pages completed"
 
-echo "[$SCRIPT_NAME] Autogenerating settings completed"
+# perform cleanup
+rm -rf clickhouse
+rm -rf "$file"
```
gingerwizard, line 96: nit:

yarn.lock
gingerwizard, line 1: why?

gingerwizard: ah rsync

scripts/copy-clickhouse-repo-docs.sh
```diff
@@ -0,0 +1,143 @@
...
+main() {
+  parse_args "$@"
+
+  # Check if rsync is available
+  has_rsync=false
+  if command -v rsync &> /dev/null; then
+    echo "rsync found"
+    has_rsync=true
+  else
+    echo "rsync not found - falling back to use cp command"
```
gingerwizard, line 119: modify to include "For faster dev builds we recommend installing rsync"

scripts/cleanup-after-build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+rm -rf scripts/tmp
```
gingerwizard, line 3: nit:

gingerwizard (changes requested):
looking good. minor changes.
Why is the build failing ???

scripts/copy-clickhouse-repo-docs.py
```diff
@@ -0,0 +1,65 @@
...
+
+def copy_docs_locally(local_path):
+  validate_local(local_path) # throw an exception if path is invalid
+  with open('package.json', 'r') as package_json:
+    package = json.load(package_json)
+    docs_folders_en = package["config"]["prep_array_en"].split()
+    docs_folders_other = package["config"]["prep_array_root"].split()
+    files_for_autogen_settings = package["config"]["autogen_needed_files"]
+  print(f"Copying docs from {local_path} ...")
+  try:
```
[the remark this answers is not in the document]
Blargian: Vercel doesn't recognise the command

Blargian:
Okay, i'm not crazy: https://stackoverflow.com/questions/68636358/file-not-found-in-alpine-container-despite-existing https://stackoverflow.com/questions/66963068/docker-alpine-executable-binary-not-found-even-if-in-path

Seems to be an underlying reason why no matter what I try I cannot run the clickhouse binary in the docs build (we just never picked this up because there was zero error handling in the original autogenerate-settings script)

Blargian:
@gingerwizard ready for review on this. I think I know why we have "new-build". 

here we pass arg "/output_path/build":

https://github.com/ClickHouse/ClickHouse/blob/0bfda2c21d37fdc82a9bff9ccadade7fe66e4d63/tests/ci/docs_check.py#L123

This is passed as an argument to `yarn build` which is just `docusaurus build` in package.json:

https://github.com/ClickHouse/clickhouse-docs/blob/2060b8d11cadab31dfd5030fbc5527f1766c3f43/package.json#L10

If I try change `yarn build` to `new-build` which calls `build-swagger` at the end of it then the docs build fails because `/output_path/build` is getting passed as an argument to `build-swagger` which then fails because it is not expecting this argument.

When we remove htmltest we'll be able to put everything in `new-build` into `build`. For now `build` needs to stay separate.

Blargian:
Need to fix broken links in settings first: https://github.com/ClickHouse/ClickHouse/pull/75524

gingerwizard (approved):

PR #3175, opened 2025-01-30
Fix scroll position of main content when left sidebar requires scrolling to show active item

DamianMaslanka5:
## Summary

Try to fix scroll position of main content when left sidebar requires scrolling to show active item

Also now the search bar is always visible, previously search bar and the first item from the sidebar was hidden after scrolling the main content.

You can see both issues on https://clickhouse.com/docs/en/optimize/skipping-indexes

I made minimum amount of changes which fixed described issues.

I tested:
1. Scrolling the left sidebar menu to active item works correctly - https://clickhouse.com/docs/en/optimize/skipping-indexes
1. Scrolling to headings works correctly - https://clickhouse.com/docs/en/sql-reference/functions/string-functions#trim

## Before

https://github.com/user-attachments/assets/84cad7d3-0754-4845-bfc6-e1a7fc165470

## After
https://github.com/user-attachments/assets/38acca53-9207-494a-8b84-c8b675aa5a84

### DIFF
--- a/src/css/custom.scss
+++ b/src/css/custom.scss
@@ -63,8 +63,7 @@ body {
 }
 /* Rich's attempt to add scrollbar to nav */
 .sidebar_mhZE {
-  position: absolute !important;
-  padding-bottom: 33px;
+  padding-bottom: 4rem;
   height: 100% !important;
 }
 /* Make the navbar move with the page */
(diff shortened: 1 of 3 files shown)


DamianMaslanka5:
@gingerwizard I thinks this is ready for a review now, I found some time browse the docs with these changes and didn't see any issues.

Blargian (approved):
Looks good to me, thanks @DamianMaslanka5!

PR #3225, opened 2025-02-07
Add a note that RO services can run CREATE/DROP USER queries

kapustor:
## Summary
Changing the warehouse page to reflect the RO services users creation/drop ability.

## Checklist


docs/en/cloud/reference/warehouses.md
```diff
@@ -129,6 +129,8 @@ Sometimes it is useful to restrict write access to a specific service and allow
 
 _Fig. 6 - Read-write and Read-only services in a warehouse_
 
+Please note - read-only services allow creation of users. This behavior can be changed in the future, however.
```
sichenzhao, line 132: creation/drop as well

PR #3228, opened 2025-02-07
Modify style check workflow to show the failed step in red

Blargian:
## Summary

When style check fails it is not immediately obvious what the issue is unless you know to expand the steps in the workflow to check which one is failing.

- spell check and kb checker now run independently and show up separately so it is clearer what failed. (also show up red or green)

## Checklist

### DIFF
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -28,7 +28,11 @@ jobs:
       - name: Run spellcheck
         run: |
           ./scripts/check-doc-aspell
+          if [ $? -ne 0 ]; then  # Check exit code immediately
+            echo "::error::KB spell check failed. See logs for details."
+          fi
         id: spellcheck
+        continue-on-error: true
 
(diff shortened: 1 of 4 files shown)


.github/workflows/check-build.yml
```diff
@@ -14,47 +14,37 @@ on:
...
+          exit_code=$?
+          elif [[ "${{ matrix.check_type }}" == "kbcheck" ]]; then
+            ./scripts/knowledgebase-checker/knowledgebase_article_checker.py --kb-dir="knowledgebase"
+          exit_code=$?
           fi
           
+          if [[ $exit_code -ne 0 ]]; then
+            echo "::error::${{ matrix.check_type }} check failed. See logs for details."
+            exit 1
+          fi
```
gingerwizard, line 50: nit:

gingerwizard (changes requested):
minor. I think i need to see what this looks like when it runs but it looks like an improvement

gingerwizard (approved):

PR #3233, opened 2025-02-09
Docs check: check for existence of ClickHouse at root, otherwise pull from master

Blargian:
On main repo I opened PR https://github.com/ClickHouse/ClickHouse/pull/75787 which attempts to align desk structure to page structure for 2 files, however docs check fails on "duplicate routes found". This is because the docs check pulls from master so when `yarn prep-from-local /ClickHouse` is run in the docker file it copies the two files (to the new directory they are moved to), and then when `yarn build` runs later in the dockerfile it pulls from master where the two files are still in the old location. This is not the desired behaviour.

This PR:

- Modifies the `copy-clickhouse-repo-docs.sh` to first attempt to look for a folder `/ClickHouse` if no `-l` command is provided. If it does not find the ClickHouse folder at the root then it will pull from master.

The logic being that:
- If we are running docs check we should use the local copy at root (pulled from the branch docs check is running on in the docker file [here](https://github.com/ClickHouse/ClickHouse/blob/449ad2d1eb17f77d4fe17c4ecadde54fe4fd0bc7/docker/docs/builder/run.sh#L7-L9))
- If we are building on vercel, we should pull the docs from master

This should allow us to be able to move files around on main repo, **except** for when a sidebar points to a specific file (i.e does not specify the menu item using a slug) because then docusaurus will complain that it can't find the file in that folder anymore. If we use automatically generated menus we won't have this problem.
 
## Checklist

### DIFF
--- a/scripts/copy-clickhouse-repo-docs.sh
+++ b/scripts/copy-clickhouse-repo-docs.sh
@@ -29,10 +29,10 @@ function parse_args() {
 # Validate local path exists and ends in ClickHouse
 function validate_local() {
 
-  if [[ -d "$local_path" && "$local_path" == *"ClickHouse" ]]; then
+  if [[ -d "$local_path" ]]; then
     return 0
   else
-    echo "Please provide a valid path to your local ClickHouse repository."
+    echo "Could not find ClickHouse repository folder at provided path: $local_path"
     exit 1
   fi
 }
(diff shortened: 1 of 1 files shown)


gingerwizard (changes requested):
discussed with @Blargian and we will just skip the pull from master if CI=true. This is less intrusive and should mean less complexity.

Blargian:
Closing in favour of https://github.com/ClickHouse/clickhouse-docs/pull/3263

PR #3252, opened 2025-02-12
fix html in search

gingerwizard:
## Summary

iframes and divs show in search drop down. We should probably remove all html from the markdown - ill follow up.

## Checklist


Blargian (approved):

DamianMaslanka5:
@gingerwizard I think using something like Beautiful Soup to parse html might be easier to maintain and less error-prone than parsing markdown with regex.

gingerwizard:
Yes aware. Was quick fix as I want to evaluate effect on relevancy of removing all html - hence comment. We measure changes in content and shift in ndcg vs perfect results

PR #3263, opened 2025-02-14
Build scripts: don't pull repo if CI environment variable is true

Blargian:
## Summary

- Needed for edge case where you make a change on main repo but the docs check fails because the copy-script is pulling from master instead of using the changes from the current branch.

## Checklist


gingerwizard (approved):

Blargian:
Require for fixing failing docs check in https://github.com/ClickHouse/ClickHouse/pull/76134

PR #3279, opened 2025-02-18
Update BYOC VPC Peering with Security Group change

yiyang-shao:
## Summary

Add an extra step to configure security group when the user intends to setup VPC Peering in BYOC
## Checklist


duanshiqiang (approved):
LGTM

Issue #3269, opened 2025-02-14
Quick Start could be more detailed


bertday:
Hello CH team! Thank you for this awesome project 😄 

I'm a brand new user and just trying out the [Quick Start](https://clickhouse.com/docs/en/getting-started/quick-start). I was surprised when I ran `clickhouse server` and it created a number of files and folders in my current working directory. It was actually a bit of a pain because I had to clean them up and then decide on where to keep my ClickHouse config. However, that decision itself was also tricky because the idea of ClickHouse config hadn't really been introduced — "where do most people run `clickhouse server` from?" is something that could have been contextualized a bit better.

I'd be happy to submit a PR for this but I'm not sure I'm quite up to speed yet. Hopefully soon 😄

Blargian:
Hi @bertday, thanks for opening this issue. You're definitely not the first one to run into that problem. When running `./clickhouse server` after the curl command the table data gets stored in the folder that the `clickhouse` binary is in. There is also a `-C [ --config-file ]` argument which can be used to set the location of the `config.xml` file that's used for configuration.  It's mentioned on the [Install](https://clickhouse.com/docs/en/install#quick-install) page but it's not so obvious. 

[Production Deployments](https://clickhouse.com/docs/en/install#available-installation-options) installs or running the prompt shown at the end of the curl command execution (`./clickhouse install`) places the configuration files here mostly: 

```
/etc/
└── clickhouse-server/
    ├── config.d/
    │   ├── data-paths.xml
    │   ├── logger.xml
    │   ├── user-directories.xml
    │   └── openssl.xml
    └── users.d/
```

There's no `./clickhouse uninstall` though, so probably I'd recommend to follow the Production Deployment if you're using ClickHouse for production otherwise just run the binary from a folder of your choosing.

I've opened a PR to update the quick-start guide to prompt the user to download the binary to a new clean subdirectory and warn them that running the server will generate some folders and files.
 
I think we want to keep the quick-start intentionally brief without diving too much into configuration details, but we should definitely link to an appropriate explainer on configuration as a next step. Maybe [here](https://clickhouse.com/docs/en/operations/configuration-files)? (This page probably needs some improvement too).

What do you think?

bertday:
I think that's a great note @Blargian ! Glad to see it added to the quick start 😄

PR #3292, opened 2025-02-19
Updates to quick-start guide

Blargian:
## Summary

Updates to quick start guide in response to https://github.com/ClickHouse/clickhouse-docs/issues/3269

## Checklist


gingerwizard (approved):

docs/en/quick-start.mdx
```diff
@@ -88,41 +126,46 @@ INSERT INTO my_first_table (user_id, message, timestamp, metric) VALUES
...
-result is a table that can be:
+Use the [`s3` table function](/docs/en/sql-reference/table-functions/s3.md) to
+read files from S3. It's a table function - meaning that the result is a table
+that can be:
 
-1. used as the source of a `SELECT` query (allowing you to run ad-hoc queries and leave your data in S3), or...
-2. insert the resulting table into a `MergeTree` table (when you are ready to move your data into ClickHouse)
+1. used as the source of a `SELECT` query (allowing you to run ad-hoc queries and
+   leave your data in S3), or...
+2. insert the resulting table into a `MergeTree` table (when you are ready to
```
gingerwizard, line 167: maybe why they would want to do this vs (1)

docs/en/quick-start.mdx
```diff
@@ -74,11 +104,19 @@ ENGINE = MergeTree
...
+You can use the familiar `INSERT INTO TABLE` command with ClickHouse, but it is
+important to understand that each insert into a `MergeTree` table causes what we
+call a **part** in ClickHouse to be created in storage. These parts later get
+merged in the background by ClickHouse.
 
-You can use the familiar `INSERT INTO TABLE` command with ClickHouse, but it is important to understand that each insert into a
-`MergeTree` table causes a **part** (folder) to be created in storage. To minimize parts, bulk insert lots of rows at a time (tens of
-thousands or even millions at once).
+In ClickHouse, we try to bulk insert lots of rows at a time
+(tens of thousands or even millions at once) to minimize the number of **parts**
```
gingerwizard, line 115: link parts to the concept guide

gingerwizard:
seems fine, minor suggestions @Blargian

docs/en/quick-start.mdx
```diff
@@ -4,63 +4,93 @@ sidebar_label: Quick Start
...
+curl https://clickhouse.com/ | sh
+```
+
+You should see:
 
-ClickHouse runs natively on Linux, FreeBSD and macOS, and runs on Windows via the [WSL](https://learn.microsoft.com/en-us/windows/wsl/about).
-The simplest way to download ClickHouse locally is to run the following `curl` command. It determines if your operating system is supported,
-then downloads an appropriate ClickHouse binary:
+```
+Successfully downloaded the ClickHouse binary, you can run it as:
```
rfraposa, line 45: This isn't a sentence....we should clean this up. Maybe

"Once the binary is downloaded, you can run it to start the server:"

docs/en/quick-start.mdx
```diff
@@ -4,63 +4,93 @@ sidebar_label: Quick Start
...
+    ./clickhouse
 
-  ```bash
-  curl https://clickhouse.com/ | sh
-  ```
+You can also install it:
```
rfraposa, line 48: Why run install? If you were doing a serious install of ClickHouse, you would use an RPM or DEB or something. And this won't even work on a Mac...

docs/en/quick-start.mdx
```diff
@@ -4,63 +4,93 @@ sidebar_label: Quick Start
...
+```
+
+You should see:
 
-ClickHouse runs natively on Linux, FreeBSD and macOS, and runs on Windows via the [WSL](https://learn.microsoft.com/en-us/windows/wsl/about).
-The simplest way to download ClickHouse locally is to run the following `curl` command. It determines if your operating system is supported,
-then downloads an appropriate ClickHouse binary:
+```
+Successfully downloaded the ClickHouse binary, you can run it as:
+    ./clickhouse
```
rfraposa, line 46: I'm not sure this makes sense in a "Quick Start" because this runs clickhouse-local, which will be confusing to someone new to ClickHouse.

docs/en/quick-start.mdx
Blargian: @rfraposa this is the output from running the curl script so it is what users will see when they follow the quick-start. Core team would need to change it.

docs/en/quick-start.mdx
Blargian: This is what's prompted in the output when you run the curl script. I added below this that we recommend **not** installing it because it indeed does not make sense to do so. There is also no `./clickhouse uninstall` command so cleaning it up is a mess.

docs/en/quick-start.mdx
Blargian: This is what gets output when you run the curl command so it's what users will see when they follow the quick-start. Agree that the prompt is confusing, but core team will need to change that if necessary.

docs/en/quick-start.mdx
gingerwizard: maybe we exclude the output given its not great and subject to change @Blargian

Blargian: @gingerwizard I can't see this changing tbh. I believe the install option was possibly added for use in light weight containers like Alpine for use in Docker, hence why there is no uninstall.

Personally I think not including it is more confusing because the user might follow the prompt to install. Better to tell them to ignore it.

Issue #3296, opened 2025-02-19
"ClickHouse Cloud" badge for ClickHouse-Cloud-only session settings


rschu1ze:
Quite a [few session settings](https://clickhouse.com/docs/en/operations/settings/settings#compatibility) only have an effect in ClickHouse Cloud but not ClickHouse OSS.

For example:

```md
distributed_cache_bypass_connection_pool

Type: Bool

Default value: 0

Only in ClickHouse Cloud. Allow to bypass distributed cache connection pool
```

or

```md
prefer_warmed_unmerged_parts_seconds

Type: Int64

Default value: 0

Only available in ClickHouse Cloud. If a merged part is less than this many seconds old and is not pre-warmed (see [cache_populated_by_fetch](https://clickhouse.com/docs/en/operations/settings/merge-tree-settings#cache_populated_by_fetch)), but all its source parts are available and pre-warmed, SELECT queries will read from those parts instead. Only for Replicated-/SharedMergeTree. Note that this only checks whether CacheWarmer processed the part; if the part was fetched into cache by something else, it'll still be considered cold until CacheWarmer gets to it; if it was warmed, then evicted from cache, it'll still be considered warm.
```

The generated docs should mark such settings with a badge, similar to the badges for experimental and beta settings.

rschu1ze:
> could we add a boolean column cloud_only to system.settings

I was afraid you would be asking for that 😄 Such a column would be simple to add of course but its presence would feel like a hidden advertisement for ClickHouse Cloud...

Suggest to go with the description approach instead. All we need to do is to standardize on some phrase, e.g. `Has an effect only in ClickHouse Cloud`.

Blargian:
> > could we add a boolean column cloud_only to system.settings
> 
> I was afraid you would be asking for that 😄 Such a column would be simple to add of course but its presence would feel like a hidden advertisement for ClickHouse Cloud...
> 
> Suggest to go with the description approach instead. All we need to do is to standardize on some phrase, e.g. `Has an effect only in ClickHouse Cloud`.

That I can understand. No problem, I'll go through and standardize the descriptions, but I worry more about new 'cloud only' settings which will get added in future with non 'standard' descriptions.

rschu1ze:
[...]

PR #3299, opened 2025-02-20
Show Cloud Only badge for session settings containing 'Only in ClickHouse Cloud' 

Blargian:
Addresses #3296 

## Summary

- Renames "core settings" -> "session settings" as per [76275](https://github.com/ClickHouse/ClickHouse/pull/76275#event-16337422743)
- Badge text shortened to "ClickHouse Cloud only" instead of "Only has an effect in ClickHouse Cloud"
- Displays the "ClickHouse Cloud only" badge if the setting description contains "Only in ClickHouse Cloud" (This is the case for the majority of them, a few need to be changed -> PR https://github.com/ClickHouse/ClickHouse/pull/76498)
## Checklist


src/theme/badges/CloudAvailableBadge/index.js
```diff
@@ -15,7 +15,7 @@ const Icon = () => {
 const CloudAvailableBadge = () => {
     return (
         <div classes className={styles.cloudBadge}>
-            <Icon />{'Only available in ClickHouse Cloud'}
+            <Icon />{'ClickHouse Cloud only'}
```
rschu1ze, line 18: Should that not read "Only has an effect in ClickHouse Cloud?"

rschu1ze: Well, I guess "ClickHouse Cloud only" is also okay (the badge applies to lots of stuff, not only settings)

rschu1ze (approved):

PR #3414, opened 2025-03-04
fix 404s + broken anchors

Blargian:
## Summary

Fixes 404s reported by HyperDX alert for 4 March
## Checklist


vercel.json
```diff
@@ -3161,6 +3161,36 @@
...
+      "destination": "/docs/interfaces/formats/Protobuf",
+      "permanent": true
+    },
+    {
+      "source": "/docs/get-started/query-endpoints",
+      "destination": "/docs/cloud/get-started/query-endpoints",
+      "permanent": true
+    },
+    {
+      "source": "/docs/getting-started/example-datasets/star-schema/u003C/p",
```
gingerwizard, line 3191: this seems unneccessary, its not a valid URL

Blargian: true but it's being linked from somewhere - 32 hits in the last 24 hours. Can't hurt to direct the traffic to the correct page.

gingerwizard: maybe but id rather the 404s be detected at source vs encouraging this by redirecting

Blargian: Fixed

PR #3432, opened 2025-03-05
Link checker throw only for locale `en`

Blargian:
## Checklist


gingerwizard (approved):

docusaurus.config.en.js
```diff
@@ -326,14 +332,6 @@ const config = {
             label: "English",
             to: "/",
           },
-          {
```
Blargian, line 329: Link checker will fail if these are here as we no longer copy across the ru and zh folders

PR #3438, opened 2025-03-06
translation improvements

gingerwizard:
## Summary

We decided to have a docusaraus config per language. Justifications:

- We want to build languages seperately to keep build times down - we also plan to deploy separately
- Each needs its own algolia and Kapa config.
- We want to deploy each under their own project with a proxy infront so /<lang>/docs makes sense.

## Checklist


scripts/build.sh
```diff
@@ -1,24 +1,26 @@
 #!/bin/bash
```
Blargian, line 1: two shebangs

Blargian (approved):

PR #3510, opened 2025-03-17
Redirects: add redirects for broken links

Blargian:
## Summary

Adds redirects/fixes for HyperDX 404 alerts
## Checklist


sidebars.js
```diff
@@ -568,33 +578,6 @@ const sidebars = {
         },
       ],
     },
-    {
```
Blargian, line 571: Removes the "Other Features" section from left sidebar of SQL reference (Introduction will get "Operators")

sidebars.js
```diff
@@ -1753,12 +1736,6 @@ const sidebars = {
           description: "Use the right table and database engines for your data",
           href: "/engines"
         },
-        {
```
Blargian, line 1756: Removes the link from the top nav to "Other features"

vercel.json
```diff
@@ -3221,6 +3226,11 @@
       "destination": "/docs/concepts/why-clickhouse-is-so-fast",
       "permanent": true
     },
+    {
```
Blargian, line 3229: Not sure where this is coming from but there is no such page. BFloat16 has always been on the same page as Floats.

gingerwizard (approved):

PR #3558, opened 2025-03-26
Russian translations V2

Blargian:
## Summary

- expands glossary
- moves prompt to `ru.json` and adds some additional prompting such as don't translate all caps, use letter ё etc.
- ~uses GPT-4O rather than GPT-4O-mini for 10 sample pages so we can explore if it's worth using~ Not a large enough difference in translation quality
- adds ability to override terms on a file level with `.translate_override`

## Notes and observations

- In V1 the prompt tells the LLM to translate yaml front matter like title, sidebar_label etc. This is in some cases not desirable as the sidebar_label could be a SQL statement or the title could be an H1.
- Still a lot of issues with frontmatter getting messed up. I think we should translate these in a more programatic way, without using an LLM.


Blargian:
https://github.com/ClickHouse/ClickHouse/pull/78338 Fix for the geo/polygon.md issue

gingerwizard (approved):

PR #3593, opened 2025-04-02
docs: add marimo as an sql client

Light2Dark:
## Summary

Hey! We're looking to add marimo as an SQL client integration, cc: @mneedham, @mscolnick

I can remove the sidebars.js change, it auto-formatted lol.

## Checklist
- [X] If adding a new integration page, also add an entry to the integrations list here: https://github.com/ClickHouse/clickhouse-docs/blob/main/docs/integrations/index.mdx


mneedham:
@Light2Dark you don't need the entry in `vercel.json`. The other entries are there because we used to have docs under `/docs/en` so the redirects are to avoid old links breaking. New docs go under `/docs` directly without the `/en` suffix, so no need to do a redirect.

mneedham:
This looks good to me 



cc @Blargian to take a look too

gingerwizard:
@Light2Dark spellchecks need to be added or fixed

gingerwizard (changes requested):
Fix spellings

Light2Dark:
I think it should be fixed, but I'm not sure how to re-check this locally. tqsm!

mneedham:
@Light2Dark I think you've fixed everything

PR #3606, opened 2025-04-06
New Best practices

gingerwizard:
## Summary

New best practices

Moving these to the main menu and re-doing.

@melvynator please read the JSON one

@tom-clickhouse for general accuracy

@Blargian for style

## Checklist


docs/best-practices/json_type.md
```diff
@@ -0,0 +1,312 @@
...
+* **Predictable nesting**: use Tuple, Array, or Nested types for these structures.
+* **Predictable structure with varying types**: consider Dynamic or Variant types instead.
+
+You can also mix approaches - for example, use static columns for predictable top-level fields and a single JSON column for a dynamic section of the payload.
+
+## Considerations and Tips for Using JSON {#considerations-and-tips-for-using-json}
+
+The JSON type enables efficient columnar storage by flattening paths into subcolumns. But with flexibility comes responsibility. To use it effectively:
+
+* **Specify path types** using [hints in the column definition](/sql-reference/data-types/newjson) to specify types for known sub columns, avoiding unnecessary type inference.
```
tom-clickhouse, line 31: Type hints not only avoid unnecessary type inference but their main advantage (not mentioned in the text) is that they completely avoid any indirection (e.g. via discriminator columns) when the data is stored and processed as JSON paths with type hints are stored exactly like traditional types or exactly as if these paths would have been extracted into separate top-level fields. That also touches the recommendation at the beginning to avoid JSON if the data is known and consistent - with type hints then there is no need to adjust ingest and schema etc - convenient and still as performant as if we would have created top level fields for all JSON paths.

docs/best-practices/json_type.md
```diff
@@ -0,0 +1,312 @@
...
+
+The JSON type enables efficient columnar storage by flattening paths into subcolumns. But with flexibility comes responsibility. To use it effectively:
+
+* **Specify path types** using [hints in the column definition](/sql-reference/data-types/newjson) to specify types for known sub columns, avoiding unnecessary type inference. 
+* **Skip paths** if you don’t need the values, with [SKIP and SKIP REGEXP](/sql-reference/data-types/newjson) to reduce storage and improve performance.
+* **Avoid setting [`max_dynamic_paths`](/sql-reference/data-types/newjson#reaching-the-limit-of-dynamic-paths-inside-json) too high** - large values increase resource consumption and reduce efficiency. As a rule of thumb, keep it below 10,000.
+
+## Advanced Features {#advanced-features}
+
+* JSON columns **can be used in primary keys** like any other columns. Codecs cannot be specified for a sub-column.
```
tom-clickhouse, line 37: > Codecs cannot be specified for a sub-column.
Not yet, but is on the roadmap, currently codecs can only be specified for the whole JSON column but not for individual paths within

docs/best-practices/minimize_optimize_joins.md
```diff
@@ -0,0 +1,64 @@
...
+* **Use direct JOINs via dictionaries if appropriate** Standard JOINs in ClickHouse are executed in two phases: a build phase which iterates the right-hand side to build a hash table, followed by a probe phase which iterates the left-hand side to find matching join partners via hash table lookups. If the right-hand side is a [dictionary](/dictionary) or another table engine with key-value characteristics (e.g. [EmbeddedRocksDB](/engines/table-engines/integrations/embedded-rocksdb) or the [Join table engine](/engines/table-engines/special/join)), then ClickHouse can use the "direct" join algorithm, which effectively removes the need to build a hash table, speeding up query processing. This works for `INNER` and `LEFT OUTER` JOINs and is preferred for real-time analytical workloads.
```
tom-clickhouse, line 35: Maybe mention that a dictionary by definition doesn't have duplicate keys and during data loading keys are silently deduplicated (last loaded key wins) - therefore for many-to-one relations if one loads the many side into a dictionary  (e.g. roles vs actors) that would result in silent data loss w.r.t the join result, many-to-mant relations not possible etc.

docs/best-practices/json_type.md
gingerwizard: Noted

Blargian (approved):

docs/best-practices/choosing_a_primary_key.md
```diff
@@ -0,0 +1,175 @@
...
+```sql
+SELECT count()
+FROM stackoverflow.posts_ordered
+WHERE (CreationDate >= '2024-01-01') AND (PostTypeId = 'Question')
+
+┌─count()─┐
+│  192611 │
+└─────────┘
+
+1 row in set. Elapsed: 0.013 sec. Processed 196.53 thousand rows, 1.77 MB (14.64 million rows/s., 131.78 MB/s.)
```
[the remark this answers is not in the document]
gingerwizard: Not sure this will work with the new widget ill add but we might need to remove

docs/best-practices/partionning_keys.md
```diff
@@ -0,0 +1,67 @@
...
+ price UInt32
+)
+ENGINE = MergeTree
+ORDER BY (town, street)
+PARTITION BY toStartOfMonth(date)
+```
+
+Whenever a set of rows is inserted into the table, instead of creating (at[ least](/operations/settings/settings#max_insert_block_size)) one single data part containing all the inserted rows (as described [here](/parts)), ClickHouse creates one new data part for each unique partition key value among the inserted rows:
+
+<Image img={partitions} size="lg" alt="Partitions" />
```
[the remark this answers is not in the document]
gingerwizard: lets pick up in image iteration

docs/best-practices/selecting_an_insert_strategy.md
```diff
@@ -0,0 +1,150 @@
...
+* The insert failed server-side and timed out.
+
+In both cases, it’s safe to **retry the insert** - as long as the batch contents and order remain identical. For this reason, it’s critical that clients retry consistently, without modifying or reordering data.
+
+### Choose the right insert target {#choose-the-right-insert-target}
+
+For sharded clusters, you have two options:
+
+* Insert directly into a **MergeTree** or **ReplicatedMergeTree** table. This is the most efficient option when the client can perform load balancing across shards. With `internal_replication = true`, ClickHouse handles replication transparently.
+* Insert into a Distributed table. This allows clients to send data to any node and let ClickHouse forward it to the correct shard. This is simpler but slightly less performant due to the extra forwarding step. `internal_replication = true` is still recommended.
```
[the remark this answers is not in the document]
gingerwizard: captial, table engine, will clarify

PR #3608, opened 2025-04-06
Add Projections page

Blargian:
## Summary

Adds a page for Projections.
- Simple example using NYC taxi dataset
- Example by Tom https://clickhouse.com/blog/clickhouse-faster-queries-with-projections-and-primary-indexes
- Moves section on ["when to use projections"](https://clickhouse.com/docs/migrations/postgresql/data-modeling-techniques#when-to-use-projections) to this doc
- We will remove the examples from [ALTER PROJECTION](https://clickhouse.com/docs/sql-reference/statements/alter/projection) page to make it more dry reference and link to this page from there.

## Checklist


docs/data-modeling/projections.md
```diff
@@ -0,0 +1,504 @@
...
+    avg(price)
+FROM uk_price_paid_with_projections
+GROUP BY county
+ORDER BY avg(price) DESC
+LIMIT 3
+SETTINGS log_queries=1
+```
+
+```response
+Query id: 7efd131f-5091-4124-89b2-2505d31ca465
```
gingerwizard, line 433: lets not include as these queries will run against sql.clickhouse

docs/data-modeling/projections.md
```diff
@@ -0,0 +1,504 @@
...
+  county,
+  price
+FROM uk_price_paid_with_projections
+WHERE town = 'LONDON'
+ORDER BY price DESC
+LIMIT 3
+SETTINGS log_queries=1
+```
+```sql
+Query id: dd2cb0ff-7082-4b71-a03c-9260d470a302 -- Note: this will be different for each user
```
gingerwizard, line 405: remove all

docs/data-modeling/projections.md
```diff
@@ -0,0 +1,545 @@
...
+    tip_amount          Float32,
+    tolls_amount        Float32,
+    total_amount        Float32,
+    payment_type        Enum('CSH' = 1, 'CRE' = 2, 'NOC' = 3, 'DIS' = 4, 'UNK' = 5),
+    pickup_ntaname      LowCardinality(String),
+    dropoff_ntaname     LowCardinality(String)
+)
+ENGINE = MergeTree
+--highlight-next-line
+PRIMARY KEY (pickup_datetime, dropoff_datetime);
```
gingerwizard, line 80: we generally don't do this but we should dicuss

docs/data-modeling/projections.md
```diff
@@ -0,0 +1,540 @@
...
+--highlight-next-line
+Processed 3.00 million rows, 
+12.12 MB (328.04 million rows/s., 1.33 GB/s.)
+Peak memory usage: 161.34 KiB.
+```
+
+Notice that because we are filtering on `tip_amount` which is not in the primary
+key ClickHouse had to do a full table scan of all 3 million rows. Although the 
+query took only 0.009 seconds to return the result, we are using only 10% of data
+in the dataset. It's therefore plausible that running this query on the full
```
gingerwizard, line 135: you mean 0.9 secs

docs/data-modeling/projections.md
```diff
@@ -0,0 +1,540 @@
...
+ALTER TABLE trips
+ADD PROJECTION prj_tip_amount
+(
+    SELECT *
+    ORDER BY tip_amount, dateDiff('minutes', pickup_datetime, dropoff_datetime)
+)
+```
+
+It is necessary after adding a projection to use the `MATERIALIZE PROJECTION` 
+statement so that the data in it is physically recalculated and ordered according
```
gingerwizard, line 151: recalculated -> "ordered and rewritten"

docs/data-modeling/projections.md
gingerwizard: Can we use the full dataset? this produces a very unrealistic example imo

docs/data-modeling/projections.md
```diff
@@ -0,0 +1,545 @@
...
+)
+ENGINE = MergeTree
+--highlight-next-line
+PRIMARY KEY (pickup_datetime, dropoff_datetime);
+```
+
+Next we'll insert 10% of the data into the table from an S3 bucket:
+
+```sql
+INSERT INTO trips SELECT
```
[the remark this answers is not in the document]
gingerwizard: lets just use the full dataset and remove this loading. Its in sql.clickhouse.com. We can take the timings from this and then just add the projection. The final query can execute against the environment (as it will have the prokjection - its just ordered by `pickup_datetime` but we can change this)

docs/data-modeling/projections.md
```diff
@@ -0,0 +1,540 @@
...
+    town LowCardinality(String),
+    district LowCardinality(String),
+    county LowCardinality(String)
+)
+ENGINE = MergeTree
+ORDER BY (postcode1, postcode2, addr1, addr2);
+```
+
+```sql title="Preprocess and insert the data"
+INSERT INTO uk_price_paid
```
gingerwizard, line 241: drop its in sql.clickhouse.com - add a note to look at the datasets on how to load the data

docs/data-modeling/projections.md
```diff
@@ -0,0 +1,540 @@
...
+
+For the original example table `uk_price_paid`, we will create (and populate) two
+projections.
+
+Let's see if we can speed this query up using Projections. To keep things tidy 
+and simple in our playground, we first duplicate the table `uk_price_paid` as 
+`uk_price_paid_with_projections`:
+
+```sql
+CREATE TABLE uk_price_paid_with_projections AS uk_price_paid;
```
gingerwizard, line 328: lets use the same technique for the taxi one, i can populate for you

docs/data-modeling/projections.md
```diff
@@ -0,0 +1,540 @@
...
+```sql
+ALTER TABLE uk_price_paid_with_projections
+  (MATERIALIZE PROJECTION prj_obj_town_price)
+SETTINGS mutations_sync = 1
+```
+
+The [`mutations_sync`](/operations/settings/settings#mutations_sync) setting is
+used to force synchronous execution.
+
+We create and populate projection prj_gby_county – an additional (hidden) table
```
gingerwizard, line 356: prj_gby_county -> `prj_gby_county`

docs/data-modeling/projections.md
```diff
@@ -0,0 +1,540 @@
...
+`AggregateFunction`. This ensures proper incremental data aggregation.
+:::
+
+The figure below is a visualization of the main table `uk_price_paid_with_projections`
+and its two projections:
+
+<Image img={projections_2} size="lg" alt="Visualization of the main table uk_price_paid_with_projections and its two projections"/>
+
+If we now run the query that lists the counties in London for the three highest 
+paid prices again, we see a dramatic difference in performance, from 0.028 sec
```
gingerwizard, line 389: LOL not quite dramatic

docs/data-modeling/projections.md
gingerwizard: also show rows scanned etc

docs/data-modeling/projections.md
```diff
@@ -0,0 +1,540 @@
...
+
+3 rows in set. Elapsed: 0.004 sec.
+```
+
+Note that both queries target the original table, and that both queries resulted
+in a full table scan (all 30.03 million rows got streamed from disk) before we 
+created the two projections.
+
+Also, note that the query that lists the counties in London for the three highest
+paid prices is streaming 2.17 million rows. When we directly used a second table
```
gingerwizard, line 442: can'#t see as you ommitted

docs/data-modeling/projections.md
```diff
@@ -0,0 +1,540 @@
...
+
+This is in contrast to Materialized Views, where the user has to select the 
+appropriate optimized target table or rewrite their query, depending on the 
+filters. This places greater emphasis on user applications and increases 
+client-side complexity.
+
+Despite these advantages, projections come with some inherent limitations which
+users should be aware of and thus should be deployed sparingly.
+
+- Projections don't allow using different TTL for the source table and the
```
gingerwizard, line 514: @tom-clickhouse i don't see anything thats out of date here but please check

gingerwizard (changes requested):

PR #3638, opened 2025-04-10
Lightweight introduction to sparse primary indexes.

tom-clickhouse:
Introduces ClickHouse’s sparse primary index and how it helps efficiently skip unnecessary data during query execution. Explains how the index is built and used, with examples and tools for observing its effect. Links to a deep dive for advanced use cases and best practices.


Blargian (approved):
@tom-clickhouse I think Dale prefers to have the responses without the line numbering. I'm fine to keep them as `output_format_pretty_row_numbers ` is enabled by default. Defer to @gingerwizard.

Code blocks don't need `runnable=true`, `runnable` works as well.

Blargian:
> Line numbers in results help a lot with following explanations as I can refer to them like in `Note how row 13 of the EXPLAIN output above shows that ...`

Agree, in this case it is needed

PR #3670, opened 2025-04-14
Revert "Add FixedString to PostgreSQL mapping"

pamarcos:
## Summary
Revert https://github.com/ClickHouse/clickhouse-docs/pull/3489 due to https://github.com/ClickHouse/ClickHouse/pull/79108

PostgreSQL's CHARACTER type stores number of characters, while ClickHouse's FixedString stores number of bytes. Thus, that 1:1 mapping doesn't work for all encodings. That's the reason it wasn't done for MySQL. Since those types mean different things, I think it's rather misleading to map one to the other. The safest option is to map CHARACTER types that mean characters to String instead.

e.g. there might be a Japanese word of 4 characters but that uses 18 bytes

## Checklist

### DIFF
--- a/docs/migrations/postgres/appendix.md
+++ b/docs/migrations/postgres/appendix.md
@@ -166,8 +166,7 @@ The following table shows the equivalent ClickHouse data types for Postgres.
 | `BIGINT` | [Int64](/sql-reference/data-types/int-uint) |
 | `SERIAL` | [UInt32](/sql-reference/data-types/int-uint) |
 | `BIGSERIAL` | [UInt64](/sql-reference/data-types/int-uint) |
-| `TEXT` | [String](/sql-reference/data-types/string) |
-| `CHAR, BPCHAR` | [FixedString](/sql-reference/data-types/fixedstring) |
+| `TEXT, CHAR, BPCHAR` | [String](/sql-reference/data-types/string) |
 | `INTEGER` | Nullable([Int32](/sql-reference/data-types/int-uint)) |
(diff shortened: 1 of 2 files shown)


pamarcos:
@Blargian any idea why the Deployment for Vercel failed? Is there any way to re-trigger that?

Blargian:
@pamarcos the translation sites are not deployed on every commit to main so they get out of sync. I can see it's failing on an unrelated change. Actually Vercel should not even try to deploy the Russian translation docs unless a change was made to those docs in this PR, so it seems something is not working as it should. Will take a look.

Merging this.

Blargian (approved):

pamarcos:
> @pamarcos the translation sites are not deployed on every commit to main so they get out of sync. I can see it's failing on an unrelated change. Actually Vercel should not even try to deploy the Russian translation docs unless a change was made to those docs in this PR, so it seems something is not working as it should. Will take a look.
> 
> Merging this.

Thanks! :hugs: 
I did change the Russian doc slightly, though
image

Blargian:
> > @pamarcos the translation sites are not deployed on every commit to main so they get out of sync. I can see it's failing on an unrelated change. Actually Vercel should not even try to deploy the Russian translation docs unless a change was made to those docs in this PR, so it seems something is not working as it should. Will take a look.
> > Merging this.
> 
> Thanks! 🤗 I did change the Russian doc slightly, though image

Woops, missed that. No problem, I'll re-run translations for that page in a follow up PR.

PR #3682, opened 2025-04-15
mvs and processing order

gingerwizard:
## Summary

Documents known behaviors.

## Checklist

### DIFF
--- a/docs/materialized-view/incremental-materialized-view.md
+++ b/docs/materialized-view/incremental-materialized-view.md
@@ -1047,7 +1047,7 @@ ORDER BY uuid ASC
 3 rows in set. Elapsed: 0.004 sec.
 ```
 
-Conversely, consider if set insert a row with `parallel_view_processing=1` enabled. With this, the views are executed in parallel, giving no guarantees with respect to the order of delivery to the target table:
+Conversely, consider what happens if we insert a row with `parallel_view_processing=1` enabled. With this enabled, the views are executed in parallel, giving no guarantees to the order at which rows arrive to the target table:
 
 ```sql
 TRUNCATE target;
--- a/docs/materialized-view/incremental-materialized-view.md
+++ b/docs/materialized-view/incremental-materialized-view.md
@@ -1182,7 +1182,7 @@ LIMIT 10
 Peak memory usage: 989.53 KiB.
 ```
 
-In ClickHouse, CTEs are inlined they are effectively copy-pasted into the query during optimization and **not** materialized. This means:
+In ClickHouse, CTEs are inlined which means they are effectively copy-pasted into the query during optimization and **not** materialized. This means:
 
 - If your CTE references a different table from the source table (i.e., the one the Materialized View is attached to), and is used in a `JOIN` or `IN` clause, it will behave like a subquery or join, not a trigger.
 - The Materialized View will still only trigger on inserts into the main source table, but the CTE will be re-executed on every insert, which may cause unnecessary overhead, especially if the referenced table is large.
--- a/docs/materialized-view/incremental-materialized-view.md
+++ b/docs/materialized-view/incremental-materialized-view.md
@@ -954,7 +954,7 @@ As shown in the previous example, a table can act as the source for multiple Mat
 
 By default, this setting is equal to `0` (`false`), meaning Materialized Views are executed sequentially in `uuid` order.
 
-For example, consider the following `source` table and 3 Materialized Views, each sending to a `target` table:
+For example, consider the following `source` table and 3 Materialized Views, each sending rows to a `target` table:
 
 ```sql
 CREATE TABLE source
--- a/docs/materialized-view/incremental-materialized-view.md
+++ b/docs/materialized-view/incremental-materialized-view.md
@@ -1083,7 +1083,7 @@ Enabling `parallel_view_processing=1` can significantly improve insert throughpu
 - **Need for strict execution order**: In rare workflows where the order of view execution matters (e.g., chained dependencies), parallel execution may lead to inconsistent state or race conditions. While possible to design around this, such setups are fragile and may break with future versions.
 
 :::note Historical defaults and stability
-Sequential execution was the default for a long time, in part due to error handling complexities. Historically, a failure in one Materialized View could prevent others from executing. Newer versions have improved this by isolating failures per block, but sequential execution still provides clearer failure semantics
+Sequential execution was the default for a long time, in part due to error handling complexities. Historically, a failure in one Materialized View could prevent others from executing. Newer versions have improved this by isolating failures per block, but sequential execution still provides clearer failure semantics.
 :::
 
 In general, enable `parallel_view_processing=1` when:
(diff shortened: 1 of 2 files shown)


docs/materialized-view/incremental-materialized-view.md
```diff
@@ -947,3 +947,255 @@ GROUP BY UserId
...
+│ 2024-03-27 │ Answer   │ 1.4907539118065434 │          703 │          0 │
+└────────────┴──────────┴────────────────────┴──────────────┴────────────┘
+
+10 rows in set. Elapsed: 0.013 sec. Processed 11.45 thousand rows, 663.87 KB (866.53 thousand rows/s., 50.26 MB/s.)
+Peak memory usage: 989.53 KiB.
+```
+
+In ClickHouse, CTEs are inlined they are effectively copy-pasted into the query during optimization and **not** materialized. This means:
+
+- If your CTE references a different table from the source table (i.e., the one the Materialized View is attached to), and it's used in a `JOIN` or `IN` clause, it will behave like a subquery or join, not a trigger.
```
Blargian, line 1187:
```suggestion
- If your CTE references a different table from the source table (i.e., the one the Materialized View is attached to), and is used in a `JOIN` or `IN` clause, it will behave like a subquery or join, not a trigger.
```

docs/materialized-view/incremental-materialized-view.md
```diff
@@ -947,3 +947,255 @@ GROUP BY UserId
...
+│ 2024-03-28 │ Question │ 1.3530864197530865 │          405 │      24564 │
+│ 2024-03-27 │ Question │ 1.3225806451612903 │          434 │      21346 │
+│ 2024-03-27 │ Answer   │ 1.4907539118065434 │          703 │          0 │
+└────────────┴──────────┴────────────────────┴──────────────┴────────────┘
+
+10 rows in set. Elapsed: 0.013 sec. Processed 11.45 thousand rows, 663.87 KB (866.53 thousand rows/s., 50.26 MB/s.)
+Peak memory usage: 989.53 KiB.
+```
+
+In ClickHouse, CTEs are inlined they are effectively copy-pasted into the query during optimization and **not** materialized. This means:
```
Blargian, line 1185:
```suggestion
In ClickHouse, CTEs are inlined which means they are effectively copy-pasted into the query during optimization and **not** materialized. This means:
```

docs/materialized-view/incremental-materialized-view.md
```diff
@@ -947,3 +947,255 @@ GROUP BY UserId
...
+        WHEN 2 THEN 'Answer'
+    END AS PostType,
+    count() AS PostsCreated,
+    avgState(Score) AS AvgScore,
+    sum(ViewCount) AS TotalViews
+FROM filtered_posts
+GROUP BY Day, PostTypeId;
+```
+
+While the CTE is strictly unnecessary here, and for example purposes, the view will work as expected:
```
Blargian, line 1146:
```suggestion
While the CTE is strictly unnecessary here, for example purposes, the view will work as expected:
```

docs/materialized-view/incremental-materialized-view.md
```diff
@@ -947,3 +947,255 @@ GROUP BY UserId
...
+
+### When to use parallel processing {#materialized-views-when-to-use-parallel}
+
+Enabling `parallel_view_processing=1` can significantly improve insert throughput, as shown above, especially when multiple Materialized Views are attached to a single table. However, it’s important to understand the trade-offs:
+
+- **Increased insert pressure**: All Materialized Views are executed simultaneously, increasing CPU and memory usage. If each view performs heavy computation or joins, this can overload the system.
+- **Need for strict execution order**: In rare workflows where the order of view execution matters (e.g., chained dependencies), parallel execution may lead to inconsistent state or race conditions. While possible to design around this, such setups are fragile and may break with future versions.
+
+:::note Historical defaults and stability
+Sequential execution was the default for a long time, in part due to error handling complexities. Historically, a failure in one Materialized View could prevent others from executing. Newer versions have improved this by isolating failures per block, but sequential execution still provides clearer failure semantics
```
Blargian, line 1086:
```suggestion
Sequential execution was the default for a long time, in part due to error handling complexities. Historically, a failure in one Materialized View could prevent others from executing. Newer versions have improved this by isolating failures per block, but sequential execution still provides clearer failure semantics.
```

docs/materialized-view/incremental-materialized-view.md
```diff
@@ -947,3 +947,255 @@ GROUP BY UserId
...
+3 rows in set. Elapsed: 0.004 sec.
+```
+
+Although our ordering of the arrival of rows from each view is the same, this is not guaranteed - as illustrated by the similarity of each row's insert time. Also note the improved insert performance.
+
+### When to use parallel processing {#materialized-views-when-to-use-parallel}
+
+Enabling `parallel_view_processing=1` can significantly improve insert throughput, as shown above, especially when multiple Materialized Views are attached to a single table. However, it’s important to understand the trade-offs:
+
+- **Increased insert pressure**: All Materialized Views are executed simultaneously, increasing CPU and memory usage. If each view performs heavy computation or joins, this can overload the system.
```
Blargian, line 1082: We should decide on "joins" vs JOINs". I prefer the latter.

docs/materialized-view/incremental-materialized-view.md
```diff
@@ -947,3 +947,255 @@ GROUP BY UserId
...
+┌─name─┬─uuid─────────────────────────────────┐
+│ mv_3 │ ba5e36d0-fa9e-4fe8-8f8c-bc4f72324111 │
+│ mv_1 │ b961c3ac-5a0e-4117-ab71-baa585824d43 │
+│ mv_2 │ e611cc31-70e5-499b-adcc-53fb12b109f5 │
+└──────┴──────────────────────────────────────┘
+
+3 rows in set. Elapsed: 0.004 sec.
+```
+
+Conversely, consider if set insert a row with `parallel_view_processing=1` enabled. With this, the views are executed in parallel, giving no guarantees with respect to the order of delivery to the target table:
```
Blargian, line 1050:
```suggestion
Conversely, consider what happens if we insert a row with `parallel_view_processing=1` enabled. With this enabled, the views are executed in parallel, giving no guarantees to the order at which rows arrive to the target table:
```

docs/materialized-view/incremental-materialized-view.md
```diff
@@ -947,3 +947,255 @@ GROUP BY UserId
...
 1 row in set. Elapsed: 0.006 sec.
 ```
+
+## Parallel vs sequential processing {#materialized-views-parallel-vs-sequential}
+
+As shown in the previous example, a table can act as the source for multiple Materialized Views. The order in which these are executed depends on the setting [`parallel_view_processing`](/operations/settings/settings#parallel_view_processing).
+
+By default, this setting is equal to `0` (`false`), meaning Materialized Views are executed sequentially in `uuid` order.
+
+For example, consider the following `source` table and 3 Materialized Views, each sending to a `target` table:
```
Blargian, line 957:
```suggestion
For example, consider the following `source` table and 3 Materialized Views, each sending rows to a `target` table:
```

PR #3692, opened 2025-04-17
New JSON guide

gingerwizard:
## Summary

Updates the JSON guide for the new type.

@Blargian for language

@melvynator please check contents

## Checklist


docs/integrations/data-ingestion/data-formats/json/schema.md
```diff
@@ -435,229 +356,320 @@ SELECT id, nickname FROM people
...
 
-The following rules can be applied to determine the most appropriate.
+More specifically, use the JSON type when your data:
 
-1. If the objects are highly dynamic, with no predictable structure and contain arbitrary nested objects, users should use the `String` type. Values can be extracted at query time using JSON functions as we show below.
-2. If the object is used to store arbitrary keys, mostly of one type, consider using the `Map` type. Ideally, the number of unique keys should not exceed several hundred. The `Map` type can also be considered for objects with sub-objects, provided the latter have uniformity in their types. Generally, we recommend the `Map` type be used for labels and tags, e.g. Kubernetes pod labels in log data.
+* Has **unpredictable keys** that can change over time.
+* Contains **values with varying types** (e.g., a path might sometimes contain a string, sometimes a number).
+* Requires schema flexibility where strict typing isn't viable.
+* You have **hundreds or even thousands** of paths which are static but simply not realistic to declare explicitly. This tends to be a rare.
```
Blargian, line 370:
```suggestion
- Has **unpredictable keys** that can change over time.
- Contains **values with varying types** (e.g., a path might sometimes contain a string, sometimes a number).
- Requires schema flexibility where strict typing isn't viable.
- You have **hundreds or even thousands** of paths which are static but simply not realistic to declare explicitly. This tends to be a rare.
```

docs/integrations/data-ingestion/data-formats/json/schema.md
```diff
@@ -435,229 +356,320 @@ SELECT id, nickname FROM people
...
+```
+
+Given the dynamic nature of the `company.labels` column between objects, with respect to keys and types, we have several options to model this data:
+
+- **Single JSON column** - represents the entire schema as a single `JSON` column, allowing all structures to be dynamic beneath this.
+- **Targeted JSON column** - only use the `JSON` type for the `company.labels` column, retaining the structured schema used above for all other columns.
+
+While the first approach [does not align with previous methodology](#static-vs-dynamic-json), a single JSON column approach is useful for prototyping and data engineering tasks. 
+
+For production deployment of ClickHouses at scale, we recommend being specific with structure and using the JSON type for targeted dynamic sub-structures where possible.
```
Blargian, line 484:
```suggestion
For production deployments of ClickHouse at scale, we recommend being specific with structure and using the JSON type for targeted dynamic sub-structures where possible. 
```

docs/integrations/data-ingestion/data-formats/json/schema.md
```diff
@@ -435,229 +356,320 @@ SELECT id, nickname FROM people
...
+
+A strict schema has a number of benefits:
+
+- **Data validation** – enforcing a strict schema avoids the risk of column explosion, outside of specific structures. 
+- **Avoids risk of column explosion** - Although the JSON type scales to potentially thousands of columns, where subcolumns are stored as dedicated columns, this can lead to a column file explosion where an excessive number of column files are created that impacts performance. To mitigate this, the underlying [Dynamic type](/sql-reference/data-types/dynamic) used by JSON offers a [`max_dynamic_paths`](/sql-reference/data-types/newjson#reading-json-paths-as-sub-columns) parameter, which limits the number of unique paths stored as separate column files. Once the threshold is reached, additional paths are stored in a shared column file using a compact encoded format, maintaining performance and storage efficiency while supporting flexible data ingestion. Accessing this shared column file is, however, not as performant. Note, however, than the JSON column can be used with [type hints](#using-type-hints-and-skipping-paths). "Hinted" columns will deliver the same performance as dedicated columns.
```
Blargian, line 489:
```suggestion
- **Avoids risk of column explosion** - Although the JSON type scales to potentially thousands of columns, where subcolumns are stored as dedicated columns, this can lead to a column file explosion where an excessive number of column files are created that impacts performance. To mitigate this, the underlying [Dynamic type](/sql-reference/data-types/dynamic) used by JSON offers a [`max_dynamic_paths`](/sql-reference/data-types/newjson#reading-json-paths-as-sub-columns) parameter, which limits the number of unique paths stored as separate column files. Once the threshold is reached, additional paths are stored in a shared column file using a compact encoded format, maintaining performance and storage efficiency while supporting flexible data ingestion. Accessing this shared column file is, however, not as performant. Note, however, that the JSON column can be used with [type hints](#using-type-hints-and-skipping-paths). "Hinted" columns will deliver the same performance as dedicated columns.
```

docs/integrations/data-ingestion/data-formats/json/schema.md
```diff
@@ -669,73 +681,93 @@ CREATE TABLE people
...
+ }
+}
+
+2 rows in set. Elapsed: 0.003 sec.
 ```
 
-This can be modelled with a `Map(String, Tuple(name String, time DateTime))` as shown below:
+### Using type hints and skipping paths {#using-type-hints-and-skipping-paths}
+
+Type hints allow us to specify the type for a path and its sub-column, thus unnecessary type inference. Consider the following example where we specify the types for the JSON keys `dissolved`, `employees`, and `founded` within the JSON column `company.labels`
```
Blargian, line 770:
```suggestion
Type hints allow us to specify the type for a path and its sub-column, preventing unnecessary type inference. Consider the following example where we specify the types for the JSON keys `dissolved`, `employees`, and `founded` within the JSON column `company.labels`
```

docs/integrations/data-ingestion/data-formats/json/schema.md
```diff
@@ -744,46 +776,190 @@ CREATE TABLE people
...
+        "type": "Array(Nullable(String))"
+ }
+}
 
-1 row in set. Elapsed: 0.001 sec.
+2 rows in set. Elapsed: 0.003 sec.
 ```
 
-The application of maps in this case is typically rare, and suggests that the data should be remodelled such that dynamic key names do not have sub-objects. For example, the above could be remodelled as follows allowing the use of `Array(Tuple(key String, name String, time DateTime))`.
+Additionally, we can skip paths within JSON that we don't want to store with [SKIP and SKIP REGEXP](/sql-reference/data-types/newjson) parameters in order to minimize storage and avoid unnecessary inference on unneeded paths. For example, suppose we use a single JSON column for the above data. We can skip the `address` and `company` paths:
```
Blargian, line 839:
```suggestion
Additionally, we can skip paths within JSON that we don't want to store with the [`SKIP` and `SKIP REGEXP`](/sql-reference/data-types/newjson) parameters in order to minimize storage and avoid unnecessary inference on unneeded paths. For example, suppose we use a single JSON column for the above data. We can skip the `address` and `company` paths:
```

docs/integrations/data-ingestion/data-formats/json/schema.md
```diff
@@ -744,46 +776,190 @@ CREATE TABLE people
...
+
+1 row in set. Elapsed: 0.450 sec.
+
+INSERT INTO people FORMAT JSONAsObject
+{"id":2,"name":"Analytica Rowe","username":"Analytica","address":[{"street":"Maple Avenue","suite":"Apt. 402","city":"Dataford","zipcode":"11223-4567","geo":{"lat":40.7128,"lng":-74.006}}],"phone_numbers":["123-456-7890","555-867-5309"],"website":"fastdata.io","company":{"name":"FastData Inc.","catchPhrase":"Streamlined analytics at scale","labels":{"type":["real-time processing"],"founded":2019,"dissolved":2023,"employees":10}},"dob":"1992-07-15","tags":{"hobby":"Running simulations","holidays":[{"year":2023,"location":"Kyoto, Japan"}],"car":{"model":"Audi e-tron","year":2022}}}
+
+1 row in set. Elapsed: 0.440 sec.
+```
+
+Note how our columns have excluded from our data:
```
Blargian, line 860:
```suggestion
Note how our columns have been excluded from our data:
```

docs/integrations/data-ingestion/data-formats/json/schema.md
```diff
@@ -744,46 +776,190 @@ CREATE TABLE people
...
-  ]
 }
+
+2 rows in set. Elapsed: 0.004 sec.
+```
+
+
+#### Optimizing performance with type hints {#optimizing-performance-with-type-hints}  
+
+Type hits offer more than just a way to avoid unnecessary type inference - they eliminate storage and processing indirection entirely, as well as allowing [optimal primitive types](/data-modeling/schema-design#optimizing-types) to be specified. JSON paths with type hints are always stored just like traditional columns, bypassing the need for [**discriminator columns**](https://clickhouse.com/blog/a-new-powerful-json-data-type-for-clickhouse#storage-extension-for-dynamically-changing-data) or dynamic resolution during query time.
```
Blargian, line 928:
```suggestion
Type hints offer more than just a way to avoid unnecessary type inference - they eliminate storage and processing indirection entirely, as well as allowing [optimal primitive types](/data-modeling/schema-design#optimizing-types) to be specified. JSON paths with type hints are always stored just like traditional columns, bypassing the need for [**discriminator columns**](https://clickhouse.com/blog/a-new-powerful-json-data-type-for-clickhouse#storage-extension-for-dynamically-changing-data) or dynamic resolution during query time. 
```

docs/integrations/data-ingestion/data-formats/json/schema.md
```diff
@@ -744,46 +776,190 @@ CREATE TABLE people
...
+```sql
+CREATE TABLE logs
+(
+    payload JSON(max_dynamic_paths = 500)
+)
+ENGINE = MergeTree
+ORDER BY tuple();
 ```
+
+**Avoid setting this parameter too high** - large values increase resource consumption and reduce efficiency. As a rule of thumb, keep it below 10,000.  For workloads with highly dynamic structures, use type hints and SKIP parameters to restrict what's stored.
```
Blargian, line 963:
```suggestion
**Avoid setting this parameter too high** - large values increase resource consumption and reduce efficiency. As a rule of thumb, keep it below 10,000.  For workloads with highly dynamic structures, use type hints and `SKIP` parameters to restrict what's stored.
```

Blargian (approved):
Approved with changes.

PR #3761, opened 2025-05-06
TPC-H docs: Update correlated subqueries

rschu1ze:
Workaround are no longer needed due to

- https://github.com/ClickHouse/ClickHouse/pull/79600
- https://github.com/ClickHouse/ClickHouse/pull/76078


novikd (commented):
I'd consider leaving decorrelated query options with mentioning that queries must be supported since `25.5`.

PR #3768, opened 2025-05-07
Related blog component

Blargian:
## Summary

Adds a related blog component which uses the keywords on each page to search for related blog posts.

Overview: 
- adds `src/components/RelatedBlogs` which queries Algolia blogs index and displays the results
- adds `RelatedBlogs` to `src/theme/DocItem/Layout/index.js` such that it shows at the bottom of each page. If a page specifies `show_related_blogs: true`
- adds keywords to many pages as they return better results than the fallback method detailed below.

Notes:
- The component searches on keywords, title, description
- Only "Engineering" blogs currently (we should add product too though imo)
- Anything from 2023-01-01 (I know we said 18 months but we have good content on joins earlier than 18 months)

To do: 
- [ ] Be able to specify docs directories from `docusaurus.config.js` so that it's not necessary to add `show_related_blogs` to each page individually (eg. we would want to exclude all reference docs).
- [ ] Mobile styling + further styling tweaks
- [ ] Only index blogs less than 18 months old

## Checklist


gingerwizard (changes requested):
1. Lets default to false and replace the existing related blog posts which are static - if a page has current related, we'll show
2. Move the keywords to the CMS and extract with Algolia

Blargian:
@gingerwizard have made the following changes:
- don't show by default, only enabled if `show_related_blogs: true` is set
- Keywords added to CMS for Engineering blogs
- I am showing anything from 2023-01-01 as we get better results for JOINs specifically. Tom has some great blogs on those.

Can always be better, but relevancy is substantially better now I think.

gingerwizard:
@Blargian can we remove any existing related content recommendations. If you search for Related Content you will find some

gingerwizard (approved):

PR #3796, opened 2025-05-13
Fix AWS Private Endpoint option value

adri:
## Summary
`false` is not a valid value in Terraform. Fixes error:

Error: expected enforce_security_group_inbound_rules_on_private_link_traffic to be one of ["on" "off"], got false


iamKunalGupta:
Thanks for the fix @adri !

iamKunalGupta (approved):

Blargian (approved):

PR #3805, opened 2025-05-15
Support using VerticalStepper with H3

Blargian:
## Summary

VerticalStepper now supports:

```
<VerticalStepper headerLevel="h3">

### One {#one}

### two {#two}

### three {#three}

</VerticalStepper>
```
## Checklist

### DIFF
--- a/plugins/remark-custom-blocks.js
+++ b/plugins/remark-custom-blocks.js
@@ -18,6 +18,8 @@ const extractText = (nodes) => {
 const plugin = (options) => {
     const transformer = (tree, file) => {
 
+        let total_steps = 0; // Keep track of the total number of steps
+
         // Target JSX elements in the AST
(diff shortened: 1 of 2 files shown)


plugins/remark-custom-blocks.js
```diff
@@ -18,6 +18,8 @@ const extractText = (nodes) => {
 const plugin = (options) => {
     const transformer = (tree, file) => {
 
+        let total_steps = 0; // Keep track of the total number of steps
```
Blargian, line 21: Added because I need to keep track of **total** number of steps because you might have multiple of these `<VerticalStepper>` components and my hacky DOM manipulation finds elements on `step-X` so I was getting interference between components.

gingerwizard (approved):

PR #3828, opened 2025-05-21
Add server-overload to sidebar

Algunenano:
Reverts ClickHouse/clickhouse-docs#3824


Blargian:
Thanks @Algunenano. It's a little annoying to have to do it this way because of the two repos. I assume the instructions were okay, but let me know if anything can be clearer.

Algunenano:
Yes, instructions were clear to me. Thanks!

PR #3867, opened 2025-05-31
MySQL ClickPipe: Document generic MySQL and MariaDB

ilidemi:
## Summary
Adding docs on how to set up self-hosted MySQL and MariaDB to work with ClickPipes.

Versions tested:
MySQL 8.4 works
MySQL 5.7 works
MySQL 5.6 can't list tables: `ERROR 1305 (42000): FUNCTION JSON_OBJECT does not exist`
MariaDB 10.5 works
MariaDB 10.3 can't validate connection: `failed to get server version: failed to compare server version: cannot parse "10.3.39-MariaDB-1:10.3.39+maria~ubu2004-log" as semver: Invalid Semantic Version` (to be fixed by https://github.com/go-mysql-org/go-mysql/pull/1028)
MariaDB 10.2 can't validate connection, semver
MariaDB 10.1 can't connect with `tls: handshake failure`, there's no overlap between cipher suites in OpenSSL 1.0.2n and what golang accepts by default https://www.skeema.io/blog/2025/02/06/mysql57-golang-ssl/
MariaDB 10.0 doesn't have `binlog_row_image` setting
MariaDB 5.5 doesn't have `binlog_row_image` setting

MySQL version prevalence: https://www.percona.com/blog/percona-server-for-mysql-8-4-2-vs-8-0-40-comparison-of-variables-and-keywords/

MariaDB version prevalence: no nice data in corporate blogs but there's Shodan:
```
========================================
Version    Count           Percentage
-----------------------------------
5.2                44      0.0%
5.5            59,551     10.0%
10.0           12,751      2.2%
10.1           78,092     13.2%
10.2           12,612      2.1%
10.3           67,615     11.4%
10.4           49,720      8.4%
10.5           58,548      9.9%
10.6          107,092     18.1%
10.7            1,986      0.3%
10.8            1,040      0.2%
10.9            1,040      0.2%
10.10           1,117      0.2%
10.11          94,086     15.9%
11.0              546      0.1%
11.1            2,603      0.4%
11.2            2,544      0.4%
11.3            3,042      0.5%
11.4            9,127      1.5%
11.5            3,519      0.6%
11.6           19,421      3.3%
11.7            6,514      1.1%
11.8              158      0.0%
-----------------------------------
Total         592,768   100.0%
```


docs/integrations/data-ingestion/clickpipes/mysql/source/generic.md
```diff
@@ -0,0 +1,139 @@
...
+
+After changing the settings, continue on with [configuring a database user](#configure-database-user).
+
+### MySQL 5.7 {#binlog-v5-x}
+
+To enable binary logging on your MySQL 5.7 instance, ensure that the following settings are configured:
+
+```sql
+server_id = 1            -- or greater; anything but 0
+log_bin = ON
```
ilidemi, line 60: MariaDB docs [strongly recommend](https://mariadb.com/kb/en/replication-and-binary-log-system-variables/#log_bin) we also set `log_bin_basename` to not be derived from hostname because that can change across time and computers. It doesn't seem like MariaDB or MySQL 8.x actually use hostname but MySQL 5.7 does. Do we need to add one more setting to the list?

serprex: Doesn't seem like a setting managed services would expose. Shouldn't matter if people use GTID like we'd like them to

ilidemi: This would technically only help in a situation where someone updates the hostname on the same machine and the file just gets truncated from the CDC perspective. But it makes sense that if you're doing that kind of sysadmin things can break, so probably no need to complicate the onboarding guide for this corner case

serprex (approved):

Blargian (approved):

PR #3890, opened 2025-06-04
New mobile menu implementation

Blargian:
## Summary

As delicious as double burger menu sounds, it should not be in a navigation menu... 
new implementation of the mobile menu on docs.
- Single burger menu
- Search is moved outside of the menu
- We show both the main menu and the side bar in the same menu
- Various tweaks and tuning to css to make the site more responsive

## Checklist


DamianMaslanka5:
FYI, [kubernetes.io](https://kubernetes.io/docs/tasks/tools/) also has two menus, IMO it looks good

image

Issue #3271, opened 2025-02-16
Improvement: auto focus search


Blargian:
**Issue**: 

When using the Algolia search popup, the search bar does not focus automatically. 

**Desired behaviour**: 

the search input should automatically focus when the pop up opens. 

**steps to reproduce**:

- Open the ClickHouse docs page using a new Chrome tab (https://clickhouse.com/docs)
- Press CMD+K (this opens the Algolia search popup). However, the text input box in the popup is not automatically focused.
- Manually click on the text input box.
- Type in a string to filter the content.

gneo0:
I would love to help solve that issue. Would you like assigning it to me?

Blargian:
Hey @gneo0, appreciate you picking this up! The handler for the search is here I believe: 

https://github.com/ClickHouse/clickhouse-docs/blob/608e46475c26e32175ebac2ce8094173f497f27a/src/theme/SearchBar/index.js#L22-L33

It's also used here for the knowledge base page search, but without an event handler.

https://github.com/ClickHouse/clickhouse-docs/blob/608e46475c26e32175ebac2ce8094173f497f27a/src/theme/BlogSidebar/Desktop/index.js#L34

gneo0:
Hey @Blargian , the bug doesn't appear when user clicks on search bar but on `Ctrl + K` keydown event happens and only at the initial render of the clickhouse docs website. The problem is that I can't locate where on the code this keydown event happens. Thanks for your provided answer and I'm looking forward if you could help me further where `Ctrl + K` event happens!

gneo0:
Hmm, great point @Blargian. So this bug should happen from Algolia? What's your thinking about that??

Blargian:
> Hmm, great point [@Blargian](https://github.com/Blargian). So this bug should happen from Algolia? What's your thinking about that??

Possible but I'm thinking more likely something wrong with the way we are using the component. DocSearch is quite commonly used so I would imagine this issue would be raised already if it was an algolia bug.

Issue #3191, opened 2025-02-02
[BUG] Using ctrl+k on knowledge base pages opens two modals


DamianMaslanka5:
https://github.com/user-attachments/assets/8ce0defb-cad4-448c-827c-f78498938362

Blargian:
Seems solved with updated knowledge base design - could not replicate on chrome and firefox.

DamianMaslanka5:
Now I can reproduce the issue on all pages.

You need to reload a page to reproduce this, it doesn't work correctly only when opening the search for the first time.

https://github.com/user-attachments/assets/a495c025-2174-48bf-b84a-3ef762e609bf

PR #3909, opened 2025-06-07
Fix showing two modals after pressing ctrl+k or / to open the search modal for the first time

DamianMaslanka5:
Current behavior is very annoying, so I decided to just fix this.

closes https://github.com/ClickHouse/clickhouse-docs/issues/3191, closes https://github.com/ClickHouse/clickhouse-docs/issues/3271

mp4 from https://github.com/ClickHouse/clickhouse-docs/issues/3191#issuecomment-2646572280 (this a bit old, now there are only two modals instead of three)

https://github.com/user-attachments/assets/a495c025-2174-48bf-b84a-3ef762e609bf

### DIFF
--- a/src/theme/SearchBar/index.js
+++ b/src/theme/SearchBar/index.js
@@ -17,6 +17,7 @@ import { useEffect } from 'react';
 import { getGoogleAnalyticsUserIdFromBrowserCookie } from '../../lib/google/google'
 
 let DocSearchModal = null;
+let searchContainer = null;
 
 function Hit({ hit, children }) {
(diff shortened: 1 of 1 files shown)


DamianMaslanka5:
@Blargian I decided to go with 2\. from https://github.com/ClickHouse/clickhouse-docs/pull/3909#issuecomment-2954866251. I rebased the branch and force pushed the changes.

Changing the SearchBar component is better because now we can be sure that another change to the layout will not reintroduce the issue.

Blargian (approved):
LGTM!

PR #3943, opened 2025-06-14
upgrade to Docusaurus 3.8.1

Blargian:
## Summary

- Upgrades Docusaurus to 3.8.1

https://docusaurus.io/blog/releases/3.8

## Checklist


docusaurus.config.en.js
```diff
@@ -46,14 +46,8 @@ const config = {
...
   future: {
-    experimental_faster: {
-      swcJsLoader: true,
-      swcJsMinimizer: true,
-      swcHtmlMinimizer: true,
-      lightningCssMinimizer: true,
-      rspackBundler: true,
-      mdxCrossCompilerCache: true,
-    },
+    experimental_faster: true,
```
Blargian, line 49: with all flags on we can replace it with a single command. `v4: true` for Docusaurus v4 future flags. See https://docusaurus.io/blog/releases/3.8#future-flags

Blargian:
Idea to test: run this PR in Amazon Linux 2023 which is what Vercel is running

PR #3985, opened 2025-06-24
Change lowercase SQL keywords to uppercase to improve consistency

DamianMaslanka5:
I wrote some code to extract sql from code blocks and used [sqruff](https://github.com/quarylabs/sqruff) to lint and fix sql, see https://github.com/DamianMaslanka5/sqruff-md.   
The code is WIP, the `fix` command works only when length of formatted sql doesn't change. I just wanted to learn some rust and check if it would be easy to extract the sql + run sqruff on extracted sql.

These changes were made automatically by running `cargo run -- fix --paths "/git/clickhouse-docs/docs/**/*.md"`

Config for sqruff is in https://github.com/DamianMaslanka5/sqruff-md/blob/master/config.cfg.

I reviewed the changes manually, I don't see any obvious issues. 

Looks like there is no existing linter for sql embedded in markdown, but it might be useful to introduce a sql linter in CI to make the formatting more consistent.

* Rules available in sqruff - https://github.com/quarylabs/sqruff/blob/main/docs/rules.md
* sqlfluff, the same as sqruff, more popular, written in python, slower - https://docs.sqlfluff.com/en/stable/reference/rules.html

### DIFF
--- a/docs/best-practices/using_data_skipping_indices.md
+++ b/docs/best-practices/using_data_skipping_indices.md
@@ -140,7 +140,7 @@ LIMIT 1
 A simple analysis shows that `ViewCount` is correlated with the `CreationDate` (a primary key) as one might expect - the longer a post exists, the more time it has to be viewed.
 
 ```sql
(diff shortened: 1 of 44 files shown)


Blargian (approved):
LGTM!

PR #4000, opened 2025-06-27
Update kafka ClickPipe setup

kschlarman:
## Summary
- Adding VerticalStepper to make it easier to read
- Updated screenshots
- Sized all images to `md` instead of `lg`
- Removed a few screenshots & simplified some text
- Added section to configure RPEs (points to private link docs)
- Extracted and revised section on schema registries

https://clickhouse-docs-5nd9brjz7-clickhouse.vercel.app/docs/integrations/clickpipes/kafka#creating-your-first-kafka-clickpipe

## Checklist


kschlarman:
@Blargian Yep, that sounds great.

Blargian (approved):

docs/integrations/data-ingestion/clickpipes/kafka.md
```diff
@@ -33,73 +28,91 @@ You have familiarized yourself with the [ClickPipes intro](./index.md).
...
+- Upstash
 
-<Image img={cp_success} alt="Success notice" size="sm" border/>
+ClickPipes is not currently compatible with the AWS Glue Schema registry or the Azure Schema Registry.
 
-<Image img={cp_remove} alt="Remove notice" size="lg" border/>
+### Configuration {#schema-registry-configuration}
 
-  The summary table provides controls to display sample data from the source or the destination table in ClickHouse
+A schema registry can be configured by when setting up a ClickPipe. This can be configured in one of three ways:
```
genzgd, line 95: "configured" twice in this line is a bit awkward.  Maybe something like "ClickPipes with Avro data require a schema registry.  This can be configured in one of three ways."

docs/integrations/data-ingestion/clickpipes/kafka.md
```diff
@@ -33,73 +28,91 @@ You have familiarized yourself with the [ClickPipes intro](./index.md).
...
+ 
+1. Providing the root schema registry URL (e.g. `https://registry.example.com`). **This is the preferred method.**
```
genzgd, line 97: Actually this is not the preferred method, because it leads to ugly UI.  Without an explicit schema, we can't display it on the next page (because we haven't gotten any data yet.). 

Screenshot 2025-07-01 at 2 45 48 PM

I would not list it first.  I would think that using the subject is the best, but I don't know what data our customers normally have access to.

genzgd: I would consider reordering this flow to get data first before displaying the schema at some point.

docs/integrations/data-ingestion/clickpipes/kafka.md
```diff
@@ -179,17 +192,6 @@ Nullable types in Avro are defined by using a Union schema of `(T, null)` or `(n
 
 ClickPipes does not currently support schemas that contain other Avro Unions (this may change in the future with the maturity of the new ClickHouse Variant and JSON data types).  If the Avro schema contains a "non-null" union, ClickPipes will generate an error when attempting to calculate a mapping between the Avro schema and Clickhouse column types.
```
genzgd, line 193: This is no longer true.  We now support non-null unions if the variant type is enabled.

genzgd (changes requested):
Sorry I missed the Avro unions part previously. Otherwise see comments.

PR #4039, opened 2025-07-08
add confluent cloud

mshustov:
## Summary
- Add a page for Fully managed ClickHouse Connect Sink on Confluent Cloud 
- Keep the old page for integrating via custom connectors for self-managed Confluent Platform 

## Checklist

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -814,7 +814,7 @@ const sidebars = {
           items: [
             "integrations/data-ingestion/kafka/index",
             "integrations/data-ingestion/kafka/kafka-clickhouse-connect-sink",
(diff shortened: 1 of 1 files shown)


Blargian:
> What is `sentence-style capitalization`? @Blargian

@mshustov it's where only the first word in the title and any nouns like product names, company names have the first letter capitalised.

E.g 

> How to connect to ClickHouse Cloud

rather than

> How To Connect To ClickHouse Cloud

Every contributor has their own different way of capitalising, so that check is just aimed at consistency. If you ever need to add an exception it can be done here: https://github.com/ClickHouse/clickhouse-docs/blob/main/styles/ClickHouse/Headings.yml

PR #4058, opened 2025-07-10
4032 spacing after list items

dhtclk:
## Summary
Addressing simple UI bug that is present on our getting started cloud page, and also documenting the edge case because it can be subtle. Started a known issues doc and linked it in the style guide for future iteration.

### DIFF
--- a/contribute/style-guide.md
+++ b/contribute/style-guide.md
@@ -161,9 +161,32 @@ Processed 335.87 thousand rows,
 13.54 MB (12.91 million rows/s., 520.38 MB/s.)
 ```
 
+### List Items
+
+List items should be properly indented so they become part of the preceeding list item as intended.
+
+For example, rather than:
+
+```text
+1. Item one
+Additional Item one Element
+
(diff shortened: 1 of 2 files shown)


Blargian:
@dhtclk This occurs on many other pages in the docs too e.g https://clickhouse.com/docs/faq/operations/production

Screenshot 2025-07-10 at 21 08 34

adding `<br/>` after the list is more of a work around. We should be able to style the elements https://github.com/ClickHouse/clickhouse-docs/tree/main/src/css so that it's not necessary to add a `<br/>` afterwards

Blargian:
For the most part looks great. There are some instances where the spacing between the list end and code block looks too much, I think because we still have the `<br/>` elements in there. Maybe grep through and remove the them now that the styling is fixed? 

Screenshot 2025-07-11 at 13 09 26
(https://clickhouse-docs-4s4zjovwx-clickhouse.vercel.app/docs/getting-started/quick-start/cloud)

PR #4113, opened 2025-07-19
Fix operators link in the sidebar menu

DamianMaslanka5:
Previously it was not possible to open the Operators page through the sidebar menu item.

https://clickhouse.com/docs/sql-reference

firefox_9gyy8kavFj


Blargian (approved):
LGTM, thanks @DamianMaslanka5!

PR #4137, opened 2025-07-20
Menu: make sidebar menus collapsible

Blargian:
## Summary

Sidebar menu is really overwhelming with the number of items. If we have a more logical structure users will not battle to find where to go and we don't have to show everything.
## Checklist

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -49,8 +49,8 @@ const sidebars = {
       type: "category",
       label: "Concepts",
       className: "top-nav-item",
-      collapsed: false,
-      collapsible: false,
+      collapsed: true,
+      collapsible: true,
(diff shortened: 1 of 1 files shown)


Blargian:
@dhtclk what are your thoughts on collapsing menu items by default to help reduce the cognitive load on users? 

See: https://clickhouse-docs-rnr1vb3j8-clickhouse.vercel.app/docs/use-cases
Screenshot 2025-07-20 at 21 28 30

Another small improvement we can make is to remove the obnoxiously large padding between menu items. See for example the Docusaurus default menu item styling:

Screenshot 2025-07-20 at 21 29 56

My personal feeling is that if content is organised very logically, having the top level items collapsed will be more useful to the user than if they are all expanded.

PR #4148, opened 2025-07-22
Update upgrades.md

krithika-ch:
Added a note
:::note
You can change release channels at any time. However, in certain cases, the change will only apply to future releases. For example, if your service is already on the Fast release channel and has received the upgrade, switching to a regular or slow release channel will not downgrade your service to a previous version. it will  follow the channel sepecific release schedule for upcoming updates. :::

## Checklist


marlenepereira (approved):
LGTM!

PR #4164, opened 2025-07-24
3966 update glossary

dhtclk:
## Summary
Adding custom tooltips that have been injected via script, based off of a json file generated off of the current glossary page. This was executed on concepts and getting-started pages only as a first pass to reduce scope.

## Checklist

https://github.com/user-attachments/assets/633cef0e-b657-4f71-b892-34d6d283a881


Blargian:
Wow! This looks super cool! 👏

I've just been looking through and realised there is one challenge to this I hadn't considered before, which is that when we run the translations the LLM is not going to know what to do with this component in the markdown. It will translate the text around it and we will be left with the translation and glossary words still in English, so by introducing this component around the words we're "polluting" the markdown a little bit.

I also wonder how this affects LLMs as they parse the markdown?
This is something we could maybe ask the folks at Kapa. We should make sure that our Ask AI is not going to spit out the component name, mistaking it as part of the text.

One potential solution is that we insert these components only when we build the docs. We should be able to do this using the Docusaurus life cycle hooks but then processing becomes trickier.

What if we introduced some custom markdown syntax like `^^ ^^` for glossary terms? E.g `A data ^^part^^ is ...`. Then at build time the custom docusaurus plugin running on the hook will transform ^^ ^^ into `<Glossary/>`. This way we don't change the text too much. I can point to some code we have already that does some of this.

Thoughts?

dhtclk:
@Blargian - You bring up some really good points. I think this is something that we need to address, I'll look into transforming custom markdown syntax, great idea.

dhtclk:
@Blargian - Updates have been made for the new approach for tool tips, along with added steps to extract the current glossary terms and check if terms are appropriately wrapped in the docs. This will only surface a warning and will not fail the build. You'll see something like this on success. 

image

Blargian:
Also realised we need to roll this out for https://clickhouse.com/docs/managing-data/core-concepts, this is actually the main place it is useful.

Blargian (approved):
LGTM

Blargian:
@dhtclk could we disable it for headers? 
Screenshot 2025-08-04 at 21 43 58

Blargian:
@dhtclk Does this have to run on `yarn start`? It should be possible to use one of the docusaurus hooks that only runs on build.

```
● Client ██████████████████████████████████████████████████ (52%) building /Users/sstruw/Desktop/clickhouse-docs/node_modules/@docusaurus/lqip-loader/lib/index.js!/Users/sstruw/Desktop/clickhouse-docs/node_modules/@docusaurus/responsive-loader/lib/index.js??ruleSet[1].rules[0].use[1]!/Users/sstruw/Desktop/
● Server ██████████████████████████████████████████████████ (53%) building /Users/sstruw/Desktop/clickhouse-docs/node_modules/@docusaurus/mdx-loader/lib/index.js??ruleSet[1].rules[8].use[0]!/Users/sstruw/Desktop/clickhouse-docs/docs/use-cases/AI_ML/MCP/ai_agent_libraries/openai-agents.md                   Processed 8 glossary terms in: cloud.mdx
Processed 28 glossary terms in: merges.mdx
Processed 115 glossary terms in: academic_overview.mdx
Processed 21 glossary terms in: partitioning_keys.mdx
Processed 19 glossary terms in: primary-indexes.mdx
```

PR #4198, opened 2025-07-31
4187 improve algolia search infra

dhtclk:
## Summary
I adjusted the Algolia Indexing rules (for the dev env) and added an ask AI button that links to our Kapa chat bot widget.


Blargian:
@dhtclk I seem to be getting an error when I try the Ask AI from the search: 

Screenshot 2025-08-01 at 11 48 32

Checked on ClickHouse.com and the same question returns results so I don't think it's a temporary service issue. Can you check if you get the same?

PR #4236, opened 2025-08-06
Allow code blocks to import code from files

Blargian:
## Summary

Per request allow code snippets to allow code imported from a file.

This change will allow the user to specify `file=code_snippets/ClickStack/config-unstructured-logs-with-processor.yaml` on a codeblock and have the code from the file inserted into the code block at build time. The markdown file itself is not actually updated but we can leave a 'stale' version of the code there for readability sake, or just leave it empty.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "node-fetch": "^3.3.2",
     "numeral": "^2.0.6",
     "prism-react-renderer": "^2.4.1",
+    "raw-loader": "^4.0.2",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -63,6 +63,7 @@
     "node-fetch": "^3.3.2",
     "numeral": "^2.0.6",
     "prism-react-renderer": "^2.4.1",
+    "raw-loader": "^4.0.2",
```
gingerwizard, line 66: what does this do?

gingerwizard: do we need?

Blargian:
@gingerwizard Could I ask for a review on this when you have time?

PR #4237, opened 2025-08-06
Add a command to build swagger from local OpenAPI

ilidemi:
## Summary
Add a command to build swagger from locally hosted OpenAPI, to easily test changes end to end when updating the spec.

## Checklist


ilidemi:
Turns out there's an action for this in the other repo already. Verified that the result is the same as in docs.

PR #4344, opened 2025-08-27
Glossary: modifications to glossary system

Blargian:
## Summary

Some small improvements:
- Expand to work with `.md` or `.mdx`
- Don't wrap terminology which is in headers (e.g. `## What are parts` should not become `## What are ^^Parts^^`
- Don't run the processor on every file in the docs, only on the folders which we ran `wrap-glossary-terms.py` on
- Fix README.md
- Remove `^^ ^^` from some parts of the docs which don't need to be wrapped (Let's begin with core-concepts, quick-starts), we can expand in future.

Future improvements:

- If a page contains x amount of glossary terms, only mark one per paragraph. I.e

```
This is a made up paragraph about ^^parts^^, which contains a lot of references to ^^parts^^. I mean like really a lot of references to ^^parts^^. So many ^^parts^^. All the ^^parts^^. ClickHouse has the best ^^parts^^.
```

could just be 

```
This is a made up paragraph about ^^parts^^, which contains a lot of references to parts. I mean like really a lot of references to parts. So many parts. All the parts. ClickHouse has the best parts.
```

We'll need to add a preprocessing step to remove `^^ ^^` before translating, or figure out a way to make this multilingual. 

## Checklist


plugins/glossary-transformer.js
```diff
@@ -50,8 +50,23 @@ function createGlossaryTransformer(options = {}) {
...
+    
+    const isInAllowedPath = allowedPaths.some(allowedPath => {
+      if (allowedPath.endsWith('/*')) {
+        const basePath = allowedPath.slice(0, -2);
+        return filePath?.includes(basePath);
+      }
+      return filePath?.includes(allowedPath);
+    });
+    
+    return (filePath?.endsWith('.mdx') || filePath?.endsWith('.md')) &&
```
Blargian, line 67: @dhtclk correct me if i'm wrong but I don't see why this won't work on .md files?

plugins/glossary-transformer.js
```diff
@@ -50,8 +50,23 @@ function createGlossaryTransformer(options = {}) {
   }
   
   function shouldProcess(filePath, fileContent) {
-    return filePath?.endsWith('.mdx') && 
+    const allowedPaths = [
+      'docs/concepts/*',
+      'docs/managing-data/core-concepts/*',
+      'docs/getting-started/*'
+    ];
```
Blargian, line 57: We seem to be processing all the files when only a small portion have `^^ ^^` so let's just do a subset. We can add as we go along.

PR #4390, opened 2025-09-05
Add new allocation profiling page to sidebar

antonio2368:
## Checklist

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -1250,7 +1250,16 @@ const sidebars = {
         "operations/ssl-zookeeper",
         "operations/startup-scripts",
         "operations/storing-data",
-        "operations/allocation-profiling",
+        {
+          type: "category",
+          label: "Allocation profiling",
+          collapsed: true,
+          collapsible: true,
+          link: { type: "doc", id: "operations/allocation-profiling" },
+          items: [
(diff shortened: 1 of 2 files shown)


sidebars.js
```diff
@@ -1250,7 +1250,16 @@ const sidebars = {
...
-        "operations/allocation-profiling",
+        {
+          type: "category",
+          label: "Allocation profiling",
+          collapsed: true,
+          collapsible: true,
+          link: { type: "doc", id: "operations/allocation-profiling" },
+          items: [
+            "operations/allocation-profiling-old",
+          ]
```
Blargian, line 1261:
```suggestion
          link: { type: "doc", id: "operations/allocation-profiling" },
          items: [
            "operations/allocation-profiling",
            "operations/allocation-profiling-old",
          ]
```

Otherwise only the old page shows up in the drop down items. I think users might not think to click the category to see the new page.

Screenshot 2025-09-05 at 13 45 45

Blargian (approved):

sidebars.js
Blargian: The new page will show as the default one. If they need the old page it's there as well below, think this is okay.
Screenshot 2025-09-05 at 14 30 07

PR #4276, opened 2025-08-14
Spanish translations

serivt:
## Summary

This PR introduces initial support for Spanish localization. It includes the setup for the new language and a translation of the **Get Started / Introduction** section into Spanish.

### What’s included

- Configuration scaffolding for Spanish language support.
- Full Spanish translation of the "Get Started / Introduction" section.
- Minor adjustments to navigation to include Spanish under language selector.

### Next steps

This is an initial partial translation. Remaining sections will follow in subsequent PRs.

## Checklist
- [ ] Delete items not relevant to your PR
- [ ] URL changes should add a redirect to the old URL via https://github.com/ClickHouse/clickhouse-docs/blob/main/docusaurus.config.js
- [ ] If adding a new integration page, also add an entry to the integrations list here: https://github.com/ClickHouse/clickhouse-docs/blob/main/docs/integrations/index.mdx

PR #4392, opened 2025-09-07
Fix some spanish translations

Wachynaky:
## Summary
A few suggested changes that I think make more sense

- address sanitizer -> verificador de direcciones: Commonly used term.

- data pipeline -> data pipeline: "Pipeline de datos" isn’t commonly used. Anyone familiar with the concept will understand "data pipeline," so it's better to leave it untranslated. A literal version would be "tubería de datos," but that sounds awkward.

- data warehousing -> data warehousing: This is a widely recognized technical term (like "big data"). If we translate it, we should also translate "data lake" to "lago de datos" ,which is not commonly done. So it’s better to keep it as is, depending on the context.

- live view -> Vista en tiempo real"Vista en tiempo real": "Vista en tiempo real" is more common in data-related technical contexts. "En vivo" is more commonly used for video streaming.

- mark files -> mark files: Like "data lake," it’s a known technical term that could be misinterpreted if translated literally. Better to keep it in English to avoid ambiguity.

- sorting key -> clave de ordenación: Correct translation in this context.

### DIFF
--- a/scripts/translate/glossary.csv
+++ b/scripts/translate/glossary.csv
@@ -2,7 +2,7 @@ word,ru,jp,zh,es
 access control,контроль доступа,access control,access control,control de acceso
 access key,ключ доступа,access key,access key,clave de acceso
 adaptive join algorithms,Адаптивные алгоритмы JOIN,adaptive join algorithms,adaptive join algorithms,algoritmos de join adaptativos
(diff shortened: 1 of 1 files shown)


Wachynaky:
I meant to include [this commit ](https://github.com/ClickHouse/clickhouse-docs/pull/4276)in this pull request, but I’m not sure how to move it. Could someone point me in the right direction?

PR #4437, opened 2025-09-16
ClickPipes: split ClickPipes billing page into two separate pages

Blargian:
## Summary

Splits the ClickPipes billing page into two separate pages for each dimension.
## Checklist


morsapaes (approved):
Thanks, @Blargian! And sorry it took me a bit long to get to this. Pushed a few changes to make the `ClickPipes` section collapsible and add a note about MySQL and MongoDB, since it isn't otherwise clear why these aren't listed. cc @iskakaushik and @saisrirampur for a 👍 on the wording:

> Usage is free for MySQL and MongoDB CDC ClickPipes before reaching General Availability (GA). Customers will be notified ahead of GA launches to review and optimize their ClickPipes usage.

Once we start billing for MySQL, we can make the section truly generic. Didn't want to play with my luck any further this time to avoid breaking redirects.

PR #4445, opened 2025-09-17
Block algolia / shortcut when ask ai widget is open

dhtclk:
## Summary
A bug was reported: using the / character in ask ai input will trigger algolia search bar to pop open

## Checklist

### DIFF
--- a/src/hooks/useAskAI.js
+++ b/src/hooks/useAskAI.js
@@ -7,7 +7,6 @@ function useAskAI() {
     const [isKapaLoaded, setIsKapaLoaded] = useState(false);
 
     useEffect(() => {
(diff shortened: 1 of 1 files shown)


Blargian:
@dhtclk kind of works, but now it's not possible to type `/` inside the kapa widget. I fixed this before, not quite sure why it's back: https://github.com/ClickHouse/clickhouse-docs/pull/4025

Blargian:
Fixed with https://github.com/ClickHouse/clickhouse-docs/pull/4448. The previous fix was targetting CSS classes which changed on Kapa's side I'm guessing. I've changed it to target the placeholder text which we control ourselves and doesn't change.

PR #4468, opened 2025-09-22
search filtering by doc_type

dhtclk:
## Summary
Indexing doc_type in algolia and adding the ability to filter by document type

image
image


Blargian (approved):
LGTM for the most part - I trust that it's working locally. Only feedback I have is that in light-mode it looks a little weird, see below:

Screenshot 2025-09-22 at 21 11 18

I feel like the drop down should maybe be positioned to the right of the search bar if possible. (or alternatively as check boxes below the search bar):

Screenshot 2025-09-22 at 21 12 10

Really nice addition! :)

dhtclk:
> LGTM for the most part - I trust that it's working locally. Only feedback I have is that in light-mode it looks a little weird, see below:
> 
> Screenshot 2025-09-22 at 21 11 18
> I feel like the drop down should maybe be positioned to the right of the search bar if possible. (or alternatively as check boxes below the search bar):
> 
> Screenshot 2025-09-22 at 21 12 10
> Really nice addition! :)

I adjusted the styling the best I could based on the limitations of customizing doc search. it's quite fickle.

PR #4489, opened 2025-09-25
Added note to CHC conversion

rsickles:
Added note to CHC conversion so customers are more clear on what they should expect from stripe invoices.


rsickles:
@Blargian looks like there is an issue with how the note was formatted...

PR #4502, opened 2025-09-30
Redirects: add redirect for weird language issue

Blargian:
## Summary

Google Search Console reports many links which aren't indexed or served on Google such as:
- https://clickhouse.com/docs/zh/jp/sql-reference/aggregate-functions/reference/uniqcombined64
- https://clickhouse.com/docs/zh/en/sql-reference/functions/geo/polygons
- https://clickhouse.com/docs/zh/ru/sql-reference/aggregate-functions/reference/last_value
- https://clickhouse.com/docs/zh/en/best-practices/selecting-an-insert-strategy
etc

Adds a redirect to the correct path based on the first language code. e.g. `https://clickhouse.com/docs/zh/en/sql-reference/functions/geo/polygons` -> `- https://clickhouse.com/docs/zh/sql-reference/functions/geo/polygons`
## Checklist

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -3500,6 +3500,11 @@
       "source": "/docs/migrations/migrations",
       "destination": "/docs/integrations/migration/overview",
       "permanent": true
+    },
+    {
+      "source": "/docs/:lang1(ru|zh|jp)/:lang2(ru|zh|jp|en)/:path*",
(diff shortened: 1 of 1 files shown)


Blargian:
No longer needed. Issue was caused by some translation files having /en in the slug still. This has not been the case for some time now so don't think this is strictly needed.

PR #4512, opened 2025-10-01
Autogenerated docs: update session and format settings scripts to use system tables and n…

Blargian:
Update scripts for session settings and format settings to use system tables rather than source code.

## Checklist


Blargian:
The idea here was to generate everything from `system.functions` but realised that there isn't a way to distinguish format and session settings currently. Discussed with Robert and we can add aliases if necessary to make sure all format settings have some kind of consistent naming which can be used to extract format settings from session settings. Need to go through and make a list of these first before progressing here.

PR #4701, opened 2025-11-04
Reformat stripe invoice note in billing overview

rsickles:
Reformat stripe invoice note in billing overview


dhtclk:
@rsickles LGTM, one random point of discussion, on the billing index page it's a bit confusing clicking the billing overview link and getting taken to the marketplace billing overview page vs the actual billing overview page. I can make note of this and create an issue, but wanted to point that out. Let me know if I'm not making sense.

dhtclk:
Screenshot 2025-11-04 at 7 56 39 PM

This link brings you to market place billing, which is fine, that's what it says. But I would expect there to be a link to the "pricing overview", or what's simply labeled as overview in the sidebar.

But not a huge deal especially if we're doing a new billing overview page anyway!

PR #4768, opened 2025-11-19
Fix `hyperdx_ingest` user grants when creating an ingestion user

execreate:
## Summary
ClickHouse ingestion user needs `CREATE DATABASE` permissions, because on startup collector tries to execute `CREATE DATABASE IF NOT EXISTS`. Not having `CREATE DATABASE` permissions will cause the collector to fail on startup.

## Checklist
- [x] If adding a new integration page, also add an entry to the integrations list here: https://github.com/ClickHouse/clickhouse-docs/blob/main/docs/integrations/index.mdx


execreate:
I am not sure what is the correct way to go here:

1. updating docs to have the correct grants
2. or updating the collector logic so that it does not try to execute `CREATE DATABASE IF NOT EXISTS`

Creating a PR to update the docs is easier, so here we are :) but I'd personally prefer the second option. What do you think? I can submit a PR for the second option later when I have a bit more time.

PR #4773, opened 2025-11-20
Auditing Vale Rules and Including a Report Script

dhtclk:
## Summary
Auditing Vale Rules and Including a Report Script

For now disabling sentence length and future tense rules. They're very subjective and flag too many false positives, so for now I've opted to exclude them in order to tackle other issues until we decide to either reinstate, remove, or refine them.

## Checklist

### DIFF
--- a/scripts/vale/vale-report.sh
+++ b/scripts/vale/vale-report.sh
@@ -7,7 +7,7 @@ vale docs/ \
 
 # Check if vale-report.json is valid JSON
 if ! jq empty vale-report.json 2>/dev/null; then
-    echo "❌ Error: Vale did not produce valid JSON output"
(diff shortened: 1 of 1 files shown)


styles/ClickHouse/CodeblockFences.yml
```diff
@@ -6,7 +6,7 @@ extends: existence
 message: "Instead of '%s' for the code block, use yaml, ruby, plaintext, markdown, javascript, shell, go, python, dockerfile, or typescript."
 link: https://docs.gitlab.com/development/documentation/styleguide/#code-blocks
 vocab: false
-level: warning
```
Blargian: This one I think we should keep as a warning personally, just because there is no syntax highlighting without it

PR #3666, opened 2025-04-13
kafka table engine, modernize table creation, add tip on limitations,…

randomizedcoder:
## Summary
Minor improvements to the kafka table engine page
- Tips on limitations of the kafka table engine ( no kafka schema, and no protobufList )
- Modernize the table creation
- Add links to full table creation options
- Add extra debug command

## Checklist
- [X] URL changes should add a redirect to the old URL via https://github.com/ClickHouse/clickhouse-docs/blob/main/docusaurus.config.js

PR #4894, opened 2025-12-03
Improve Kafka Table Engine documentation

morsapaes:
Something I had stashed for a while, but never finished. Needs a few more iterations, which I hope I'll finally get to over the holidays.


randomizedcoder:
Great!  Thanks for the update. - We should mention this thread started in https://github.com/ClickHouse/clickhouse-docs/pull/3666

It would be great to mention the batching that Kafka engine performs.  It's now been many months since I looked at this, but from memory it batches for 15 seconds by default, so the inserts do happen in batching.  This is why, although protobufList doesn't work, and doesn't have test automation, this isn't really a problem, given the batching.

It would also be good to mention that Kafka schema registry is unsupported.

PR #4990, opened 2025-12-15
Persist galaxy user id across sessions and improved request handling

thelevicole:
## ID persistence

The current implementation generates a new galaxy user ID on each session and stores it in `sessionStorage`. Because `sessionStorage` is scoped to a single tab and cleared when the session ends, the same user can receive multiple user IDs across tabs and browser restarts.

Additionally, cloud signup links currently append the galaxy **session** ID, when they should instead use the Galaxy **user** ID

This PR resolves both issues by:

1. **Introducing a persistent galaxy user ID**
   - A UUID is generated once and reused, with persistence limited by client-side factors e.g. cleared browser data or cookie expiration.
   - The ID is stored and retrieved in multiple places (in order of priority):
     1. Cookie
     2. `localStorage`
     3. `sessionStorage`
   This ensures consistency across tabs and sessions while maintaining backward compatibility.

2. **Updating cloud signup links**
   - Cloud signup URLs now append the galaxy **user** ID rather than the session ID, ensuring stable user attribution.
   
## Request reliability
   
Additionaly, this PR includes improved handling of galaxy requests with beacon/fetch hybrid.

Previously, all galaxy events were sent using fetch without keepalive. This meant requests could be dropped during page unload/navigation.

Switches plain fetch to a hybrid beacon-first, fetch-fallback strategy.

1. Uses `navigator.sendBeacon` for small payloads (queued on unload).
2. Falls back to fetch (keepalive when possible, normal fetch for large payloads).
3. Adds transport metadata (_transport, _queued) for clarity.
4. Improves reliability during navigation and safer handling of large events.

### DIFF
--- a/src/lib/galaxy/galaxy.js
+++ b/src/lib/galaxy/galaxy.js
@@ -36,7 +36,7 @@ function findOrCreateGalaxyId(key) {
 }
 
 export function getUserId() {
-  // Use the cookie/storage name `glx_anonymous_id` for backwards compatability
+  // Use the cookie/storage name `glx_anonymous_id` for backwards compatibility
   return findOrCreateGalaxyId('glx_anonymous_id')
 }
 


package.json
```diff
@@ -111,4 +112,4 @@
   "resolutions": {
     "form-data": "^4.0.4"
   }
-}
```
Blargian: I'm morally obliged by @gingerwizard to ask for a new line here :)

src/lib/galaxy/galaxy.js
```diff
@@ -1,19 +1,106 @@
...
+    }
+  }
+
+  const id = read() ?? uuid()
+  persist(id)
+  return id
+}
+
+export function getUserId() {
+  // Use the cookie/storage name `glx_anonymous_id` for backwards compatability
```
Blargian: nit
```suggestion
  // Use the cookie/storage name `glx_anonymous_id` for backwards compatibility
```

PR #5002, opened 2025-12-16
Integrations: move interfaces and cpp client pages to clickhouse-docs repo

Blargian:
## Summary

Moves the following pages to this repo under `integrations/interfaces` (and `integrations/language-clients` for cpp.md)
- interfaces/cpp
- interfaces/http
- interfaces/tcp
- interfaces/jdbc
- interfaces/mysql
- interfaces/odbc
- interfaces/postgresql
- interfaces/prometheus
- interfaces/ssh
- interfaces/grpc
- interfaces/arrowflight

## Checklist

### DIFF
--- a/scripts/sed_links.sh
+++ b/scripts/sed_links.sh
@@ -24,6 +24,12 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
     sed -i '' 's|(/cloud/security/secure-s3#access-your-s3-bucket-with-the-clickhouseaccess-role)|(/cloud/data-sources/secure-s3#access-your-s3-bucket-with-the-clickhouseaccess-role)|g' docs/sql-reference/table-functions/s3.md
     sed -i '' 's|(/cloud/security/secure-s3#access-your-s3-bucket-with-the-clickhouseaccess-role)|(/cloud/data-sources/secure-s3#access-your-s3-bucket-with-the-clickhouseaccess-role)|g' docs/sql-reference/table-functions/s3Cluster.md
     sed -i '' 's|(#cuttofirstsignificantsubdomaincustom)|(#cutToFirstSignificantSubdomainCustom)|g' docs/sql-reference/functions/url-functions.md
+    # Fix broken links in interface docs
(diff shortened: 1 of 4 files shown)


scripts/sed_links.sh
```diff
@@ -24,6 +24,25 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
     sed -i '' 's|(/cloud/security/secure-s3#access-your-s3-bucket-with-the-clickhouseaccess-role)|(/cloud/data-sources/secure-s3#access-your-s3-bucket-with-the-clickhouseaccess-role)|g' docs/sql-reference/table-functions/s3.md
     sed -i '' 's|(/cloud/security/secure-s3#access-your-s3-bucket-with-the-clickhouseaccess-role)|(/cloud/data-sources/secure-s3#access-your-s3-bucket-with-the-clickhouseaccess-role)|g' docs/sql-reference/table-functions/s3Cluster.md
     sed -i '' 's|(#cuttofirstsignificantsubdomaincustom)|(#cutToFirstSignificantSubdomainCustom)|g' docs/sql-reference/functions/url-functions.md
+    # Fix broken links in interface docs
```
Blargian: This script gets run at build time when the ClickHouse/ClickHouse docs get pulled into this repo to build the docs. There's unfortunately no way to selectively exclude files from the built-in broken link checker of Docusaurus so we need to trick it here. I will open a PR on ClickHouse/ClickHouse next to remove the "old" files and then we can remove these.

plugins/floating-pages-exceptions.txt
```diff
@@ -10,4 +10,15 @@
 integrations/language-clients/java/client-v1
 integrations/language-clients/java/jdbc-v1
 integrations/data-ingestion/clickpipes/postgres/maintenance.md
-operations/backup.md
+interfaces/cpp
```
Blargian: The old files still exist but are not assigned to any sidebar now. This tricks the CI into not complaining about this.

PR #5251, opened 2026-01-19
Add `LIMIT`s to TPC-H queries 2, 3, 10, 18 and 21

rienath:
## Summary
[The document](https://www.tpc.org/TPC_Documents_Current_Versions/pdf/TPC-H_v3.0.1.pdf) that defines TPC-H benchmark does not have `LIMIT` statements in some of the queries. But in paragraph 2.1.2.9, they ask to add them for queries 2, 3, 10, 18 and 21, which we do in this PR. 

Also, a cosmetic improvement for Q15.

### DIFF
--- a/docs/getting-started/example-datasets/tpch.md
+++ b/docs/getting-started/example-datasets/tpch.md
@@ -676,33 +676,33 @@ WHERE
 **Q15**
 
 ```sql
-with revenue_view as (
-    select
-        l_suppkey as supplier_no,
-        sum(l_extendedprice * (1 - l_discount)) as total_revenue
-    from
+WITH revenue_view AS (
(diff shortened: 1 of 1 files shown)


Blargian (approved):

rschu1ze:
@rienath Would you like to check and add limits if needed for TPC-H Qs 2, 3, 10, 18, 21 in tests/private_performance (private repository)? I found a couple of query files w/o LIMIT, e.g. tests/private_performance/tpch_adapted_1_official/queries/query_10.sql - it's especially odd in this case because the `official` flavor is supposed to be the "the original TPC-H queries" (according to tests/private_performance/README.md) and I believe we report internally on these.

PR #5350, opened 2026-01-27
update search route

dhtclk:
## Summary
Due to how rspackbundler handles routing, the docusaurus config needed to be updated. this change occured when rspackBundler: true was enabled in experimental faster features.

## Checklist


Blargian (approved):
LGTM, preview deploys seem to be behaving correctly. Merging.

PR #5504, opened 2026-02-10
Remove custom webpack configuration to improve initial load time

DamianMaslanka5:
With custom configuration there were 400 http requests. With default configuration there are "only" 100 requests and the initial page load is 1s faster locally (2.5s => 1.5s), the difference should be bigger after the change is deployed to https://clickhouse.com/docs

I see improvements locally, but these changes should be first deployed to a vercel preview env to make sure everything still works correctly and similar improvements are visible when the docs are deployed to vercel

The changes with custom config were introduced in https://github.com/ClickHouse/clickhouse-docs/pull/4770, but I don't see any information why these changes are needed.

### Before (with VERCEL=1)

master

### After

default-config

cc: @Blargian @dhtclk

### DIFF
--- a/docusaurus.config.en.js
+++ b/docusaurus.config.en.js
@@ -322,42 +322,6 @@ const config = {
         },
       };
     },
-    // Webpack optimization plugin for large sites
-    function webpackOptimizationPlugin(context, options) {
-      return {
-        name: 'webpack-optimization-plugin',
-        configureWebpack(config, isServer) {
-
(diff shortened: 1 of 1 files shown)


dhtclk:
@DamianMaslanka5 - Thanks for making this change, this was originally a change to fix some very long vercel build times at the time, but it clearly no longer is providing any benefit, and will need to be removed anyway.

Once we see good vercel build times then I will merge!

PR #5534, opened 2026-02-16
[dbt] Improve Materialization View docs

koletzilla:
## Summary
- Improve title.
- Add "Troubleshooting common issues" section.
- Add "Debugging techniques" section.
- Reestructure behavior comparisions.
- Improve introduction for the behavior differences during active ingestion (related to https://github.com/ClickHouse/dbt-clickhouse/issues/612#issuecomment-3908129408)

# Other things tested and not included in the docs

##### Column incompatibilities between MV and target table -> this is printed by CH.

```
DB::Exception: SELECT query outputs column with name 'non_existing_column', which is not found in the target table. Use 'AS' to assign alias that matches a column name.
```
##### Target doesnt exists yet** -> we already provide instructions in the error log

```
Target table ' ~ materialization_target_table ~ ' not found in cache. It may not exist yet. Ensure the target table model runs before this MV.'
```
##### "Target has changed" error** -> we already provide instructions in the error log

```
'Target table mismatch for MV "' ~ mv_relation.name ~ '": Current target is "' ~ existing_target
~ '", but model references "' ~ materialization_target_table_name ~ '". '
~ 'The TO clause cannot be changed via ALTER TABLE MODIFY QUERY. '
~ 'Use --full-refresh to recreate the MV with the new target table.'
```
##### Orphaned MV warning after rename** -> We already have instructions in the warning log for the implicit target version. Explicit target is managed as other removed dbt resource, so users know it needs to be manually removed.

```
{{ log('Warning - Materialized view "' + table + '" was detected pointing to the model "' + target_relation.name + '" but was not found in this run. In case it is a renamed mv that was previously part of this model, drop it manually (!!!)', info=True) }}
```
##### Target table out of sync** -> we already provide instructions in the error log

```
The source and target schemas on this {0} model are out of sync.

They can be reconciled in several ways:
- set the `{1}` config to `append_new_columns` or `sync_all_columns`.
- Re-run the {0} model with `full_refresh: True` to update the target schema.
- update the schema manually and re-run the process.

Additional troubleshooting context:
Source columns not in target: {2}
Target columns not in source: {3}
New column types: {4}
```


docs/integrations/data-ingestion/etl-tools/dbt/materialization-materialized-view.md
```diff
@@ -278,13 +278,53 @@ To change the target:
...
+#### Target table is empty while/after `run` is executed {#target-table-empty}
+There are a few reasons why this can happen:
+- The MVs may be configured with `catchup=False` or the target table may be configured with `repopulate_from_mvs_on_full_refresh=False`, so no backfill is executed when the MVs are created or when the target table is recreated. This is expected behavior, so if you want to reinsert the data using the MVs SQL, make sure to set `catchup=True` in the MVs (this is the default value) or `repopulate_from_mvs_on_full_refresh=True` in the target table. Make sure you are not activating both at the same time to avoid duplicates. Check the [configuration section](#explicit-target-configuration) for more details.
```
BentsiLeviav: Do you see any situation where both are needed to be set? If not, should we check for that in the adapter somehow and raise a warning/exception?

docs/integrations/data-ingestion/etl-tools/dbt/materialization-materialized-view.md
```diff
@@ -278,13 +278,53 @@ To change the target:
...
+        create_table_query,
+        '.*TO\\s+`?([^`\\s(]+)`?\\.`?([^`\\s(]+)`?.*',
+        '\\1.\\2'
+    ) AS target_table
+FROM system.tables
+WHERE database = 'your_schema'
+  AND engine = 'MaterializedView'
+```
+
+#### Check if dbt recognizes a table as an MV target {#check-dbt-recognition}
```
Blargian:
```suggestion
#### Check if dbt recognizes a table as a materialized view target {#check-dbt-recognition}
```

Blargian (approved):

PR #5656, opened 2026-03-04
Restructure auto scaling docs

Blargian:
## Summary

Restructure autoscaling docs
## Checklist

### DIFF
--- a/docs/getting-started/quick-start/cloud.mdx
+++ b/docs/getting-started/quick-start/cloud.mdx
@@ -47,7 +47,7 @@ Once you're logged in, ClickHouse Cloud starts the onboarding wizard which walks
 <Image img={createservice1} size="md" alt='New ClickHouse Service' border/>
 <br/>
 
-By default, new organizations are put on the Scale tier and will create 3 replicas each with 4 VCPUs and 16 GiB RAM. [Vertical autoscaling](/cloud/features/autoscaling/vertical#vertical-auto-scaling) will be enabled by default in the Scale tier. You can change your organization tier later on the 'Plans' page.
+By default, new organizations are put on the Scale tier and will create 3 replicas each with 4 VCPUs and 16 GiB RAM. [Vertical autoscaling](/cloud/features/autoscaling/vertical) will be enabled by default in the Scale tier. You can change your organization tier later on the 'Plans' page.
 
 Customize the service resources if needed by specifying a minimum and maximum size for replicas to scale between. When ready, select `Create service`.
 


aashishkohli:
Otherwise the changes look good. We will also need to add a page to explain "Scheduled Scaling" which is a feature currently in preview but will start rolling out more broadly in 2-3 weeks.

We can add that section as "Preview" feature if that is possible? Suggested page content is below.

---

# Scheduled Scaling

ClickHouse Cloud services automatically scale based on CPU and memory utilization, but many workloads follow predictable patterns — daily ingestion spikes, batch jobs that run overnight, or traffic that drops sharply on weekends. For these use cases, Scheduled Scaling lets you define exactly when your service should scale up or down, independent of real-time metrics.

With Scheduled Scaling, you configure a set of time-based rules directly in the ClickHouse Cloud console. Each rule specifies a time, a recurrence (daily, weekly, or custom), and the target size — either the number of replicas (horizontal) or the memory tier (vertical). At the scheduled time, ClickHouse Cloud automatically applies the change, so your service is sized appropriately before demand arrives rather than reacting after the fact.

This is distinct from metric-based autoscaling, which responds dynamically to CPU and memory pressure. Scheduled Scaling is deterministic: you know exactly when the scaling will happen and to what size. The two approaches are complementary — a service can have a baseline scaling schedule and still benefit from autoscaling within that window if workloads fluctuate unexpectedly.

Scheduled Scaling is currently available in **Private Preview**. To enable it for your organization, contact the ClickHouse support team.

## Setting up a scaling schedule

To configure a schedule, navigate to your service in the ClickHouse Cloud console and go to settings. From there, select **Schedule Override** and add a new rule.

<embed screenshots here>

*The Scaling Schedules interface in the ClickHouse Cloud console, showing time-based scaling rules.*

Each rule requires:

- **Time:** When the scaling action should occur (in your local timezone)
- **Recurrence:** How often the rule repeats (e.g. every weekday, every Sunday)
- **Target size:** The number of replicas or memory allocation to scale to

Multiple rules can be combined to form a full weekly schedule. For example, you might scale out to 5 replicas every weekday at 6 AM and scale back to 2 replicas at 8 PM.

## Use cases

**Batch and ETL workloads:** Scale up before a nightly ingest job runs and scale back down once it completes, avoiding over-provisioning during idle daytime hours.

**Predictable traffic patterns:** Services with consistent peak hours (e.g. business-hours query traffic) can be pre-scaled to handle load before it arrives, rather than waiting for autoscaling to react.

**Weekend scale-down:** Reduce replica count or memory tier over weekends when demand is lower, then restore capacity before the Monday morning surge.

**Cost control:** For teams managing ClickHouse Cloud spend, scheduled scale-downs during known low-utilization periods can meaningfully reduce resource consumption without any manual intervention.

:::note
A scheduled scaling action and a concurrent autoscaling recommendation may interact — the schedule takes precedence at its trigger time.
:::

PR #5772, opened 2026-03-18
Update remote MCP doc

Blargian:
Updates in response to feedback that our remote MCP server docs could be better

Focus:
- Simplify where possible
- Extend the doc beyond just Claude, taking inspiration from [Linear's page](https://linear.app/docs/mcp#cursor) which I personally have used
- We can keep this doc as a source of truth as more first class connectors/plugins release
## Checklist

### DIFF
--- a/docs/cloud/guides/AI_ML/remote_mcp.md
+++ b/docs/cloud/guides/AI_ML/remote_mcp.md
@@ -89,7 +89,6 @@ Select `clickhouse-cloud` and authenticate via OAuth using your credentials for 
 
 ### Visual Studio Code {#visual-studio-code}
 
-
 Add the following configuration to your `.vscode/mcp.json`:
 
 ```json


docs/cloud/guides/AI_ML/remote_mcp.md
```diff
@@ -22,74 +22,130 @@ import img5 from '@site/static/images/use-cases/AI_ML/MCP/5connected_mcp_claude.
...
+Copy the displayed URL, which is the same as the one below:
 
 ```bash 
 https://mcp.clickhouse.cloud/mcp
 ```
 
-## Add the ClickHouse MCP Server in Claude Code {#add-clickhouse-mcp-server-claude-code}
+## Setup remote MCP for development {#setup-clickhouse-cloud-remote-mcp-server}
+
+Choose your IDE or tool below and follow the corresponding setup instructions.
```
dhtclk: Not sure if would be worth using the tab component here or something similar - the list is quite long so it may not be worth it.

PR #5777, opened 2026-03-18
[fivetran] Revamp docs

koletzilla:
Revamp Fivetran docs:
- `Index` cover more details about the current version and related files
- `Reference` covers all the docs we have in Fivetran side: configurations, table engine details...
  - I have not moved the full `setup-guide`, only the configurations details. The full guide should be kept complete in Fivetran side as it's really nice to have it when editing the connector. 
- `Troubleshooting` contains usual errors with possible solutions, good practices and some examples to help debugging.

### DIFF
--- a/docs/integrations/data-ingestion/etl-tools/fivetran/troubleshooting.md
+++ b/docs/integrations/data-ingestion/etl-tools/fivetran/troubleshooting.md
@@ -12,7 +12,7 @@ keywords: ['fivetran', 'clickhouse destination', 'troubleshooting', 'best practi
 
 ## Common errors {#common-errors}
 
-### Grants test failed or operations are failed related to permissions {#grants-test-failed}
+### Grants test failed or operations are failing related to permissions {#grants-test-failed}
 
 **Error message:**
 


docs/integrations/data-ingestion/etl-tools/fivetran/reference.md
```diff
@@ -0,0 +1,210 @@
...
+SETTINGS index_granularity = 8192
+```
+
+Since `_fivetran_id` is unique and there are no other primary key options, it is used as a table sorting key.
+
+### Selecting the latest version of the data without duplicates {#selecting-latest-version}
+
+`SharedReplacingMergeTree` performs background data deduplication
+[only during merges at an unknown time](/engines/table-engines/mergetree-family/replacingmergetree).
+However, selecting the latest version of the data without duplicates ad-hoc is possible with the `FINAL` keyword and
```
BentsiLeviav: Should we warn that `FINA` can be significantly slower on large tables? 
In addition, do you think we might be able to overcome duplicates using MV, not 100% sure, but something like:

```sql
  -- MV that feeds it
  CREATE MATERIALIZED VIEW analytics.orders_mv
  TO analytics.orders_latest
  AS SELECT *
  FROM fivetran_schema.orders
  WHERE _fivetran_deleted = false;
```

docs/integrations/data-ingestion/etl-tools/fivetran/troubleshooting.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Don't manually modify Fivetran-managed tables {#dont-modify-tables}
+
+Avoid manual DDL changes (e.g., `ALTER TABLE ... MODIFY COLUMN`) to tables managed by Fivetran. The connector expects the schema it created. Manual changes can cause [type mapping errors](#uint64-type-error) and schema mismatch failures.
+
+Use materialized views for custom transformations.
+
+### Ensure cluster health during syncs {#cluster-health}
+
+The Fivetran destination checks that all replicas are active before performing operations. If any replica is offline, operations fail after retrying for up to 600 seconds.
```
BentsiLeviav: IIRC that was (partially) changed. 

`WaitAllNodesAvailable` runs before mutations and is a monitoring/warning step. If replicas aren't all up, it logs a warning but does not block the operation. https://github.com/ClickHouse/clickhouse-fivetran-destination/blob/904aef0da7f0815d75032224f43f24a6ae80b6f0/destination/db/clickhouse.go#L946 

`WaitAllMutationsCompleted` is the one that actually retries for up to 600s, but only when code 341 is received after the mutation itself.

docs/integrations/data-ingestion/etl-tools/fivetran/reference.md
```diff
@@ -0,0 +1,210 @@
...
+    "select_batch_size": 200
+  }
+}
+```
+
+## Type transformation mapping {#type-mapping}
+
+The Fivetran ClickHouse destination maps [Fivetran data types](https://fivetran.com/docs/destinations#datatypes) to ClickHouse types as follows:
+
+| Fivetran type | ClickHouse type |
```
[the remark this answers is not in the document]
koletzilla: uops, yes, it's. I see some differences. Adding it to the docs

dhtclk (approved):

PR #5877, opened 2026-03-27
Add native python UDF documentation for chdb

wudidapaopao:
## Summary
Add documentation for chdb's new native Python UDF API (`@func` / `create_function`), including:
- New guide page: Python UDF Guide covering registration, type system, NULL/exception handling, and DateTime support
- Updated API reference with new UDF section and legacy API preserved with deprecation notice
- Simplified UDF section in install guide, linking to the full guide


dhtclk (approved):
LGTM, ran all of the code examples and they worked as expected.

PR #6044, opened 2026-04-14
Update ClickStack demo docs

pulpdrew:
## Summary

Update the ClickStack demo documentation based on changes to the product and to the demo data

1. Demo data covers 48 hours, so searching the last 2 days is now recommended
2. Update the images for event deltas, since the UI has been updated (also the previous description was misleading, as visa are the faster events, not the slower ones)
3. Update the suggested client session email search term, due to updated demo data

## Checklist


Blargian (approved):
LGTM

PR #6092, opened 2026-04-21
Add scale factor 10 import example to TPC-H dataset docs

rienath:
## Summary
I recently added TPC-H SF10 data to S3 to make development easier for myself. Updating docs to expose it so others don't have to generate SF10 locally via `tpch-kit` either. The page already documents SF1 (good for small tests) and SF100 (slow to pull), but nothing in between. SF10 is the mid-sized sweet spot.


rschu1ze (approved):

i18n/jp/docusaurus-plugin-content-docs/current/getting-started/example-datasets/tpch.md
```diff
@@ -79,7 +79,7 @@ tpch-kit を使用して自分でテーブルを生成する代わりに、公
 
 
 ```sql
--- Scaling factor 1
+-- Scale factor 1
```
rschu1ze, line 82: @Blargian Do we really need explicit translation pages for other languages? I suppose that makes the long-term maintenance of our docs really difficult.

Blargian: @rschu1ze there is no need to make updates to the i18n folder ever, there is an agent that listens for changes on the English language docs and opens a PR with the translations automatically.

Blargian (approved):

PR #6158, opened 2026-05-06
[kafka-connect]: Add troubleshooting entry on handling large connection counts

kurnoolsaketh:
## Summary
Adds an entry addressing an edge case in which high data volume connector deployments may result in a large number of open connections to ClickHouse.

### DIFF
--- a/docs/integrations/data-ingestion/kafka/kafka-clickhouse-connect-sink.md
+++ b/docs/integrations/data-ingestion/kafka/kafka-clickhouse-connect-sink.md
@@ -827,7 +827,7 @@ transforms.keyToValue.field=_key
 ```
 
 #### "There are too many open connections to my ClickHouse instance" {#too-many-open-connections}
-In high task count connector deployments where the insert frequency is high, insertions to ClickHouse may result in many open connections to your database. In ClickHouse Cloud, a common symptom of this issue is requests being rate limited by the cloud proxy/load balancer.
+High insertion frequency may result in many open connections to your database. This is common in large task count or distributed connector deployments where the insert rate to a single ClickHouse instance is high. In ClickHouse Cloud, a common symptom of this issue is requests being rate limited by the cloud proxy/load balancer.
 
 Some strategies to reduce the number of open connections are:
 1. Adjust the connection pool settings on the Java client (note that these may reduce overall throughput):
--- a/docs/integrations/data-ingestion/kafka/kafka-clickhouse-connect-sink.md
+++ b/docs/integrations/data-ingestion/kafka/kafka-clickhouse-connect-sink.md
@@ -825,3 +825,15 @@ transforms=keyToValue
 transforms.keyToValue.type=com.clickhouse.kafka.connect.transforms.KeyToValue
 transforms.keyToValue.field=_key
 ```
+
+#### "There are too many open connections to my ClickHouse instance" {#too-many-open-connections}
+In high task count connector deployments where the insert frequency is high, insertions to ClickHouse may result in many open connections to your database. In ClickHouse Cloud, a common symptom of this issue is requests being rate limited by the cloud proxy/load balancer.
+
+Some strategies to reduce the number of open connections are:
+1. Adjust the connection pool settings on the Java client (note that these may reduce overall throughput):
+   - `max_open_connections`: defaults to 10. Reducing this will bound the number of open connections per task.
+   - `connection_ttl`: defaults to -1 (no ttl). Setting this to >0 will eagerly reclaim connections after the ttl expires.
+
+   See the [Java client Connection & Endpoints configuration tab](https://clickhouse.com/docs/integrations/language-clients/java/client#configuration) for more details.
+
+2. Increase `bufferCount`: in high data volume/throughput deployments, this will increase the number of records buffered between inserts and reduce the frequency of insert queries sent to your database. This will reduce the number of new connections needed to write to your database.


docs/integrations/data-ingestion/kafka/kafka-clickhouse-connect-sink.md
```diff
@@ -825,3 +825,15 @@ transforms=keyToValue
 transforms.keyToValue.type=com.clickhouse.kafka.connect.transforms.KeyToValue
 transforms.keyToValue.field=_key
 ```
+
+#### "There are too many open connections to my ClickHouse instance" {#too-many-open-connections}
+High insertion frequency may result in many open connections to your database. This is common in large task count or distributed connector deployments where the insert rate to a single ClickHouse instance is high. In ClickHouse Cloud, a common symptom of this issue is requests being rate limited by the cloud proxy/load balancer.
+
+Some strategies to reduce the number of open connections are:
+1. Adjust the connection pool settings on the Java client (note that these may reduce overall throughput):
```
kurnoolsaketh, line 833: @chernser are java client options configurable on the connector?

chernser: yes, they are configured via `jdbcConnectionProperties` this is similar to JDBC url. However if configuration has some unknown property it may be treated as ClickHouse settings for V1

docs/integrations/data-ingestion/kafka/kafka-clickhouse-connect-sink.md
```diff
@@ -825,3 +825,15 @@ transforms=keyToValue
 transforms.keyToValue.type=com.clickhouse.kafka.connect.transforms.KeyToValue
 transforms.keyToValue.field=_key
 ```
+
+#### "There are too many open connections to my ClickHouse instance" {#too-many-open-connections}
```
chernser, line 829: I would name it shorter "Too many DB connections".

docs/integrations/data-ingestion/kafka/kafka-clickhouse-connect-sink.md
```diff
@@ -825,3 +825,15 @@ transforms=keyToValue
...
+High insertion frequency may result in many open connections to your database. This is common in large task count or distributed connector deployments where the insert rate to a single ClickHouse instance is high. In ClickHouse Cloud, a common symptom of this issue is requests being rate limited by the cloud proxy/load balancer.
```
chernser, line 830: This needs careful explanation because there may be many records and we are fine with that and there can be many small batches. 

Current text declares problem way too broadly and do not explain why it happens. 

It may be worth having structuring the record in way 
```
<Description> 

Symptoms on prem: 
Symptoms on cloud:
Metrics to check: 

<What configuration to change, how to troubleshoot>

File: docs/integrations/data-ingestion/kafka/kafka-clickhouse-connect-sink.md [hunk 4]
```diff
@@ -825,3 +825,15 @@ transforms=keyToValue
...
+
+Some strategies to reduce the number of open connections are:
+1. Adjust the connection pool settings on the Java client (note that these may reduce overall throughput):
+   - `max_open_connections`: defaults to 10. Reducing this will bound the number of open connections per task.
+   - `connection_ttl`: defaults to -1 (no ttl). Setting this to >0 will eagerly reclaim connections after the ttl expires.
+
+   See the [Java client Connection & Endpoints configuration tab](https://clickhouse.com/docs/integrations/language-clients/java/client#configuration) for more details.
```
chernser, line 837: Java client is not configured via endpoint. 
But troubleshooting guide expects that user know how to configure client in the connector - if it is not explained before then it should be fixed.

docs/integrations/data-ingestion/kafka/kafka-clickhouse-connect-sink.md
```diff
@@ -825,3 +825,15 @@ transforms=keyToValue
...
+
+2. Increase `bufferCount`: in high data volume/throughput deployments, this will increase the number of records buffered between inserts and reduce the frequency of insert queries sent to your database. This will reduce the number of new connections needed to write to your database.
```
chernser, line 839: Is ti really for data volume and throughput? 
This is not - before that sink was handling big batches just fine.

chernser (changes requested):

docs/integrations/data-ingestion/kafka/kafka-clickhouse-connect-sink.md
kurnoolsaketh: > Symptoms on prem:

AFAIK, we haven't seen a case of this happening in on-prem yet since this issue was originally raised by a cloud customer. Depending on if and how their LB is configured, this may not even be an issue for on-prem customers.

But, I agree with your suggestion to include `Metrics to check` - i have added it 👍 thanks!

docs/integrations/data-ingestion/kafka/kafka-clickhouse-connect-sink.md
kurnoolsaketh: this is a little confusing to me - are you saying user can set `max_open_connections` by passing it in `jdbcConnectionProperties`? This is not a JDBC property though.

docs/integrations/data-ingestion/kafka/kafka-clickhouse-connect-sink.md
kurnoolsaketh: > But troubleshooting guide expects that user know how to configure client in the connector

Agreed, I am just linking it here for ease of reference.

docs/integrations/data-ingestion/kafka/kafka-clickhouse-connect-sink.md
kurnoolsaketh: yes this is incorrect. I've removed `in high data volume/throughput deployments, `, you are right that not high data volume/tput cases will encounter this issue.

PR #6165, opened 2026-05-06
Update 01_export-backups-to-own-cloud-account.md

aashishkohli:
Removed the "Enterprise Feature" badge as the feature is now available across all tiers.

@Blargian - how can we check if there are other references to this being an "Enterprise" only feature? It's available across all tiers now so I want to make sure our docs are capturing that consistently.

## Checklist


dhtclk:
@aashishkohli - I went ahead and did some investigating with Claude code to find other areas that referenced backups as an enterprise only feature and committed the changes. Want to double check?

Blargian (approved):

PR #6233, opened 2026-05-14
Add operations/query-rules to sidebar

alexey-milovidov:
ClickHouse PR https://github.com/ClickHouse/ClickHouse/pull/88234 introduces
a new docs page at `docs/en/operations/query-rules.md` for the new Query
Rewrite Rules feature.

Without an entry in `sidebars.js` the page is reported as a "floating page"
by the docs build, which fails the docs check on the ClickHouse PR.

This adds `'operations/query-rules'` to the operations sidebar group, next to
the existing query-cache / query-condition-cache entries.

### DIFF
--- a/plugins/floating-pages-exceptions.txt
+++ b/plugins/floating-pages-exceptions.txt
@@ -13,3 +13,4 @@ integrations/data-ingestion/clickpipes/postgres/maintenance.md
 operations/utilities/clickhouse-keeper-http-api.md
 cloud/features/backups/faq.md
 interfaces/web-terminal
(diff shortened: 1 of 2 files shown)


Blargian (approved):
It has to be done in two steps as Docusaurus build's fail if there is an entry in sidebar.js pointing to a page which it cannot find:

1. Add the page to floating-pages-exception.txt, rerun docs check of ClickHouse/ClickHouse PR and merge 
2. Once merged, remove the page from the exception list and add it to sidebar.js

PR #6244, opened 2026-05-18
Snowflake -> ClickHouse Equivalent Concepts

dhtclk:
## Summary
Snowflake -> ClickHouse equivalent concepts page to strengthen our migration story.

## Checklist


docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,244 @@
...
+
+## SQL and functions {#sql-and-functions}
+
+The query-language surface: SQL coverage, UDFs, and the built-in function library.
+
+ClickHouse SQL covers the standard `SELECT` / `JOIN` / `GROUP BY` / window-function surface. Function-by-function mapping (date, JSON, string, regex, window) lives in the [Snowflake → ClickHouse SQL translation reference](/migrations/snowflake-translation-reference); the rows below are concept-level only.
+
+| Snowflake | ClickHouse | Notes |
+|---|---|---|
+| Snowflake SQL | ClickHouse SQL — same `SELECT` / `JOIN` / `GROUP BY`, with [lambdas](/sql-reference/functions/overview#arrow-operator-and-lambda) and aggregate [combinators](/sql-reference/aggregate-functions/combinators) as additional language features | Compatible at the level of basic SQL. Lambdas and combinators are the two extensions worth getting familiar with. |
```
Blargian: This is more of an implementation detail. I would expect to find this in the SQL reference page

dhtclk: Good call out, I'll remove for now.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Resource hierarchy {#resource-hierarchy}
+
+| Snowflake | ClickHouse | Notes |
+|---|---|---|
+| Organization | [Organization](/cloud/security/console-roles#organization-roles) | Root node of the hierarchy in both. |
+| Account | [Warehouse](/cloud/reference/warehouses) | Each service scales compute independently; storage is shared at the warehouse level. Tier and billing are set at the organization level, not per warehouse. |
+| Database | [Database](/sql-reference/statements/create/database) | Logical container for tables. Snowflake uses a Database → Schema → Table hierarchy; ClickHouse flattens this to Database → Table. See [Schemas](#schemas) below. |
+
+:::note Warehouse terminology
```
Blargian:
```suggestion
:::note[Warehouse terminology]
```

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Logical grouping by domain or processing stage (`analytics.raw`, `analytics.staging`, `analytics.marts`) | Separate databases or a consistent naming convention | — |
+| Permission boundary | [SQL grants](/sql-reference/statements/grant) at the database, table, or column level | Database-wide grants cover the schema-level grant footprint; per-table grants are also available for finer-grained control. |
+| Future grants | Database wildcards (`GRANT … ON db.* TO role`) apply to current and future tables | Can't scope future grants to a subset of tables within a database. |
+| Schema `OWNERSHIP` and `MANAGED ACCESS` | — | ClickHouse has no object-ownership model, so grants are always explicit. |
+| Cloning unit (`CREATE SCHEMA … CLONE`) | No copy-on-write at any granularity; see the [Storage and tables](#storage-tables) section for the zero-copy clone row | Every copy reads source data fully in ClickHouse. |
```
Blargian: Fact checking against Kapa:

ClickHouse does support zero-copy table cloning via `CREATE TABLE … CLONE AS`, which uses hard-linking and part-level copy-on-write. The claim that "every copy reads source data fully" is incorrect. The document itself seems to contradict this row by referencing a "zero-copy clone row" elsewhere. [[table cloning](https://clickhouse.com/blog/table-cloning#from_immutable_parts_to_instant_clones)]

The correct nuance is that cloning is at the table level, not the schema level — so there is no `CREATE SCHEMA … CLONE` equivalent, but zero-copy cloning does exist at the table level.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Time Travel and replication boundary | Handled at the table level ([TTL](/sql-reference/statements/create/table#ttl-expression)) or service level ([backups](/cloud/manage/backups/overview), [database replication](#operations)) | No intermediate per-schema boundary. |
```
Blargian: Partially accurate, but TTL is not the right equivalent for Time Travel. TTL in ClickHouse controls data expiry/deletion, not point-in-time data recovery.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+
+| Snowflake | ClickHouse | Notes |
+|---|---|---|
+| Account-level system roles (`ACCOUNTADMIN`, `SYSADMIN`, `SECURITYADMIN`, `USERADMIN`, `PUBLIC`) | [Organization roles](/cloud/security/console-roles#organization-roles) (Admin, Billing, Org API reader, Member) and [service roles](/cloud/security/console-roles#service-roles) (Service admin, Service reader, Service API admin/reader) in the console; SQL roles inside each service | Org-scoped console roles cover billing, org admin, and user management; service-scoped roles cover service config, scaling, and backups. |
+| Custom account roles | [`CREATE ROLE`](/sql-reference/statements/create/role) in SQL | Same pattern: create a role, grant privileges to it, grant the role to users. |
+| Database roles | — | ClickHouse has only one tier of SQL roles, all service-scoped. No equivalent to Snowflake's two-tier account/database role split. |
```
Blargian: Mostly accurate, but incomplete. The knowledge sources confirm that SQL roles in ClickHouse are service-scoped. However, ClickHouse Cloud does have SQL console-specific roles (sql_console_admin, sql_console_read_only) and supports user-level granular roles via the sql-console-role:<email> namespace pattern, which adds a layer of nuance not captured here. [[common access queries](https://clickhouse.com/docs/cloud/security/common-access-management-queries)]

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Role hierarchy (`GRANT ROLE … TO ROLE …`) | [`GRANT role1 TO role2`](/sql-reference/statements/grant) | — |
+| Privilege grants on objects (`GRANT … ON … TO ROLE …`) | [`GRANT … ON db.table TO role`](/sql-reference/statements/grant) | — |
+| Object ownership and ownership transfer | — | Access in ClickHouse is controlled entirely through explicit grants. Snowflake patterns that rely on owners delegating access need to be rebuilt as explicit role-based grants. |
+| `USE ROLE` | [`SET ROLE`](/sql-reference/statements/set-role) | — |
+
+## Compute and capacity {#compute-capacity}
+
+| Snowflake | ClickHouse | Notes |
+|---|---|---|
+| Virtual warehouse | Service (one or more replicas) | A ClickHouse service runs across one or more replicas (typically 3 by default on Scale/Enterprise); queries [parallelize](/optimize/query-parallelism) across them. |
```
Blargian: Needs clarification. The knowledge sources state that Scale/Enterprise tier services support 2+ replicas, and the Basic tier supports only single-replica services. The claim of "3 by default" is not explicitly confirmed — the sources say "2+ zones" for Scale/Enterprise availability. [[cloud tiers](https://clickhouse.com/docs/cloud/manage/cloud-tiers)]

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Query Acceleration Service | No direct equivalent | ClickHouse has no per-query compute booster; scale the service via [vertical autoscaling](/cloud/features/autoscaling/vertical) if queries are consistently large. |
+
+## Billing and pricing model {#billing}
+
+ClickHouse Cloud bills compute as RAM-minutes rather than credits scaled by warehouse size, storage as compressed bytes without Time Travel or Fail-safe overhead, and backups as their own line item rather than bundled into retention windows. Most Snowflake "serverless compute" features (Snowpipe, Search Optimization, Auto-clustering, materialized view refresh, Cortex) are bundled into service compute on ClickHouse; [ClickPipes](/integrations/clickpipes) is the explicit exception and is [metered separately](/cloud/reference/billing/clickpipes). Both platforms charge for public internet egress and cross-region data transfer, and both offer committed-spend discounts. See [ClickHouse Cloud pricing](/cloud/manage/billing/overview) for current rates, tiers, and commitment options.
```
Blargian: Let's use "compute units" rather than RAM-minutes, which would be new terminology.ClickHouse Cloud meters compute on a per-minute basis in 8 GiB RAM increments. See [[compute model](https://clickhouse.com/blog/how-cloud-data-warehouses-bill-you#clickhouse-cloud)] for details.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Stage (internal / external / user / table) | Object storage referenced directly via [`s3`](/sql-reference/table-functions/s3) / [`gcs`](/sql-reference/table-functions/gcs) / [`azureBlobStorage`](/sql-reference/table-functions/azureBlobStorage) table functions; [ClickPipes](/integrations/clickpipes) for managed staging on load | ClickHouse has no stage object: there's no managed internal storage layer for files awaiting load, and no `PUT` / `GET` equivalents for moving files in and out. Read from the bucket directly, or use ClickPipes to coordinate ingest. |
+| Iceberg table (managed or unmanaged) | [Iceberg engine](/engines/table-engines/integrations/iceberg) | Reads Iceberg tables stored in S3, Azure, HDFS, or local storage. Writes are not supported. See the engine page for the current list of supported features. |
```
Blargian: This will go stale fast - for e.g. we announced write support for Iceberg last week. Let's just refer to https://clickhouse.com/docs/use-cases/data-lake/support-matrix as the source of truth

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Secure view | [View](/sql-reference/statements/create/view) with `SQL SECURITY DEFINER` (runs with the view-owner's privileges) | See [CREATE VIEW](/sql-reference/statements/create/view) for the syntax and the `INVOKER` / `DEFINER` / `NONE` modes. |
+| Row access policy | [Row policy](/sql-reference/statements/create/row-policy) — a `WHERE`-style expression evaluated per user | Row policies apply transparently to every query against the table. |
+| Sequence | No direct equivalent — use [`generateSnowflakeID`](/sql-reference/functions/uuid-functions#generateSnowflakeID), [`generateUUIDv7`](/sql-reference/functions/uuid-functions#generateUUIDv7), or an external generator | ClickHouse has no auto-incrementing sequence object; generated IDs are produced per row at insert time. |
+
+:::note Time Travel and backups
+ClickHouse Cloud backups are per-service rather than per-table.
+Restoring creates a new service, historical state isn't queryable
+inline, and a single table can't be restored back into the original
+service.
+:::
```
Blargian: Same comment here as in the BigQuery PR - backups are per-service if taken via the UI, but you can backup individual tables using SQL BACKUP/RESTORE commands. See: https://clickhouse.com/docs/operations/backup/overview#syntax

Blargian: Let's also use the `[ ]` syntax here for the note title

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Clustering key | [`ORDER BY`](/guides/best-practices/sparse-primary-indexes) columns in the table definition | Where Snowflake's clustering key is advisory and reorganized in the background, ClickHouse's `ORDER BY` is enforced at insert time and drives the sparse primary index. |
+| Data retention (table / database default) | [`TTL` clause](/sql-reference/statements/create/table#ttl-expression) on the table, column, or partition | `TTL` automatically deletes data older than a configured window. Set at table creation or via [`ALTER TABLE ... MODIFY TTL`](/sql-reference/statements/alter/ttl). |
+| Time Travel | Point-in-time [backup](/cloud/manage/backups/overview) restore into a new service | Granularity differs significantly; see the callout below. |
+| Fail-safe | — | Recovery beyond the backup window goes through ClickHouse Cloud support, not a self-service tier. |
```
Blargian, line 98: I would confirm accuracy of this with support or with @aashishkohli

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Zero-copy clone | [`CREATE TABLE ... AS SELECT`](/sql-reference/statements/create/table) copy, or [backup](/cloud/manage/backups/overview) restore into a new service | ClickHouse has no copy-on-write primitive; every copy reads the source data fully. |
```
Blargian: ClickHouse does support zero-copy table cloning via `CREATE TABLE ... CLONE AS`, which uses hard-linking at the part level.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Secure view | [View](/sql-reference/statements/create/view) with `SQL SECURITY DEFINER` (runs with the view-owner's privileges) | See [CREATE VIEW](/sql-reference/statements/create/view) for the syntax and the `INVOKER` / `DEFINER` / `NONE` modes. |
```
Blargian: From Kapa, take with a pinch of salt:

Partially accurate. The knowledge sources confirm ClickHouse supports SQL SECURITY DEFINER, INVOKER, and NONE modes on views. [[CREATE VIEW](https://clickhouse.com/docs/sql-reference/statements/create/view#refreshable-materialized-view)]

However, Snowflake's Secure View does more than just run with the owner's privileges — it also hides the view definition from non-owners and prevents certain query optimizations that could expose underlying data. The document maps only the privilege-delegation aspect, which is incomplete. The knowledge sources do not contain enough information to confirm whether ClickHouse has an equivalent to Snowflake's definition-hiding behavior.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Row access policy | [Row policy](/sql-reference/statements/create/row-policy) — a `WHERE`-style expression evaluated per user | Row policies apply transparently to every query against the table. |
+| Sequence | No direct equivalent — use [`generateSnowflakeID`](/sql-reference/functions/uuid-functions#generateSnowflakeID), [`generateUUIDv7`](/sql-reference/functions/uuid-functions#generateUUIDv7), or an external generator | ClickHouse has no auto-incrementing sequence object; generated IDs are produced per row at insert time. |
```
Blargian: Incomplete as of v25.1. The knowledge sources confirm that ClickHouse now has generateSerialID(), a named distributed counter stored in ClickHouse Keeper, which is specifically designed for auto-incrementing sequences:

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+
+:::note Updates and deletes
+ClickHouse is append-optimized. There's no SQL `MERGE` statement
+(unrelated to the `Merge` and `MergeTree` engines), and
+[`ALTER TABLE … UPDATE`](/sql-reference/statements/alter/update) /
+[`DELETE`](/sql-reference/statements/alter/delete) run as background mutations
+rather than transactional row writes. Update patterns from Snowflake (`MERGE`,
+[dbt incremental](#transformation) updates) typically port to engine choice in ClickHouse:
+[`ReplacingMergeTree`](/engines/table-engines/mergetree-family/replacingmergetree)
+keeps the latest row by sort key, [`CollapsingMergeTree`](/engines/table-engines/mergetree-family/collapsingmergetree)
```
Blargian, line 112: For clarity: "CollapsingMergeTree cancels rows by inserting a matching row with Sign = -1."

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Sequence | No direct equivalent — use [`generateSnowflakeID`](/sql-reference/functions/uuid-functions#generateSnowflakeID), [`generateUUIDv7`](/sql-reference/functions/uuid-functions#generateUUIDv7), or an external generator | ClickHouse has no auto-incrementing sequence object; generated IDs are produced per row at insert time. |
+
+:::note Time Travel and backups
+ClickHouse Cloud backups are per-service rather than per-table.
+Restoring creates a new service, historical state isn't queryable
+inline, and a single table can't be restored back into the original
+service.
+:::
+
+:::note Updates and deletes
```
Blargian:
```suggestion
:::note[Updates and deletes]
```

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Stream (CDC over a table) | [Materialized view](/materialized-views) over base-table inserts, or [ClickPipes](/integrations/clickpipes) for source-side CDC | ClickHouse MVs react on each insert; there's no offset/consume model. |
+| `EXPLAIN` / `EXPLAIN_JSON` | [`EXPLAIN`](/sql-reference/statements/explain) variants (`PLAN`, `PIPELINE`, `SYNTAX`, `ESTIMATE`) | `EXPLAIN ESTIMATE` reports rows, parts, and marks the query would read; other variants cover deeper plan inspection. |
+| External functions | No direct equivalent — closest options are [executable UDFs](/sql-reference/functions/udf) (local script invocation) or a [database engine](/engines/database-engines) attaching a live source | ClickHouse has no managed outbound HTTP call from SQL; the surrogates run locally or attach a database, not call an arbitrary service. |
+| Sessions / session variables | Per-statement execution; multi-step state managed in the client or an orchestrator | ClickHouse has no per-session variables or shared state. |
+
+## Transformation and modeling {#transformation}
```
Blargian, line 138: Our resident dbt expert @amychen1776 can help us here :)

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Foreign key (advisory) | Wide tables or [dictionaries](/dictionary) for lookups | ClickHouse doesn't accept foreign-key declarations even as advisory hints. |
+| Search Optimization Service | Secondary indexes — [bloom-filter](/engines/table-engines/mergetree-family/mergetree#bloom-filter), token-bloom, [minmax](/engines/table-engines/mergetree-family/mergetree#minmax) | ClickHouse asks you to pick the index type per column and tune its parameters; there's no automatic equivalent. |
+| Cortex Search / Snowflake Cortex Search | [Full-text index](/engines/table-engines/mergetree-family/textindexes) | Token index over string columns for in-database search. |
+| `VECTOR` data type and vector search | [`Array(Float32)`](/sql-reference/data-types/array) plus a [vector ANN index](/engines/table-engines/mergetree-family/annindexes) | ClickHouse has no dedicated `VECTOR` type. Embeddings are stored as `Array(Float32)` and accelerated with an ANN index for approximate nearest-neighbor lookups. |
```
Blargian: Missing QBit and BFloat16 for completeness

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Task (scheduled SQL) | [Refreshable MV](/materialized-view/refreshable-materialized-view) for query-driven scheduled work; external orchestrator ([dbt](/integrations/dbt), Airflow) for procedural pipelines | Task DAGs have no direct equivalent; model dependencies in the orchestrator. |
+| Stream (CDC over a table) | [Materialized view](/materialized-views) over base-table inserts, or [ClickPipes](/integrations/clickpipes) for source-side CDC | ClickHouse MVs react on each insert; there's no offset/consume model. |
+| `EXPLAIN` / `EXPLAIN_JSON` | [`EXPLAIN`](/sql-reference/statements/explain) variants (`PLAN`, `PIPELINE`, `SYNTAX`, `ESTIMATE`) | `EXPLAIN ESTIMATE` reports rows, parts, and marks the query would read; other variants cover deeper plan inspection. |
+| External functions | No direct equivalent — closest options are [executable UDFs](/sql-reference/functions/udf) (local script invocation) or a [database engine](/engines/database-engines) attaching a live source | ClickHouse has no managed outbound HTTP call from SQL; the surrogates run locally or attach a database, not call an arbitrary service. |
```
Blargian: Let's point users specifically to https://clickhouse.com/docs/engines/table-engines/special/url - table engine which queries data to/from a remote HTTP/HTTPS server. https://clickhouse.com/docs/sql-reference/table-functions/remote is also relevant.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Sessions / session variables | Per-statement execution; multi-step state managed in the client or an orchestrator | ClickHouse has no per-session variables or shared state. |
```
Blargian: `SET setting_name = value` applies a setting for the lifetime of the current session - all subsequent queries in that session use the new value.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Other transformation frameworks (Coalesce, SQLMesh, etc.) | Use the tool's ClickHouse adapter | Adapter coverage and maturity vary; verify supported features against the tool's own documentation. |
+
+## Security and governance {#security-governance}
+
+Secure views and row access policies are listed under [Storage and tables](#storage-tables). Roles and grants are covered in [Roles and access control](#roles-access).
+
+| Snowflake | ClickHouse | Notes |
+|---|---|---|
+| Column masking policies (including tag-based) | Column-level [grants](/sql-reference/statements/grant) on specific columns of a table, or [data masking patterns](/cloud/guides/data-masking) | Grants apply at the column level. Snowflake's centralized tag/policy governance has no direct equivalent. |
+| Dynamic data masking (function-based) | Views, [row policies](/sql-reference/statements/create/row-policy), or function-based transforms; see [data masking patterns](/cloud/guides/data-masking) | No column-mask primitive; patterns are SQL-level. |
```
Blargian: Inaccurate as of ClickHouse Cloud 25.12+. `CREATE MASKING POLICY` is precisely a column-mask primitive - it applies function-based transformations to specific columns dynamically at query time for specific roles, without modifying stored data. [[CREATE MASKING POLICY](https://clickhouse.com/docs/sql-reference/statements/create/masking-policy)]

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Network policies (IP allowlist) | IP allowlists and [private connectivity](/cloud/security/connectivity/private-networking) — PrivateLink (AWS, Azure) and Private Service Connect (GCP) for ingress restriction | Private connectivity is available across the three major clouds. |
+| Tri-Secret Secure (customer-managed keys) | [CMEK](/cloud/security/cmek) on the service | Available on AWS (KMS) and GCP, with rotation and revocation. |
```
Blargian: This will go stale whenever we support Azure in future. Link to the CMEK docs rather.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Audit logs (`ACCOUNT_USAGE.LOGIN_HISTORY`, `QUERY_HISTORY`) | [Cloud audit log](/cloud/security/audit-logging) and [`system.query_log`](/operations/system-tables/query_log) | Admin events go to the audit log; query activity to `system.query_log`. |
+
+## Data sharing {#data-sharing}
+
+| Snowflake | ClickHouse | Notes |
+|---|---|---|
+| Secure Data Sharing | Read access to a shared database, or a dedicated service with consumer-specific [row policies](/sql-reference/statements/create/row-policy) | ClickHouse has no zero-copy cross-account share; sharing uses standard access primitives. |
+| Snowflake Marketplace / Listings | — | ClickHouse has no in-product data marketplace. |
+| Reader accounts (provider-managed consumer) | Dedicated service per consumer, or shared service with row policies | Consumers must have their own ClickHouse Cloud account; no equivalent for serving non-customers under the provider's billing. |
+| Data Clean Rooms | [Row policies](/sql-reference/statements/create/row-policy) and [secure views](/sql-reference/statements/create/view) | No managed clean-room product. |
```
Blargian: Important to call out that `CREATE MASKING POLICY` (Cloud 25.12+) adds dynamic column masking at query time [[masking policies](https://clickhouse.com/docs/cloud/guides/data-masking#masking-policies)]

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Operations and ecosystem {#operations}
+
+ClickHouse surfaces operational state through `system.*` tables (queries, sessions, replication, parts, metrics) and the cloud console; managed ingestion is handled by ClickPipes; ML, BI, and notebook workflows are handled in external systems that read from ClickHouse.
+
+| Snowflake | ClickHouse | Notes |
+|---|---|---|
+| Snowpipe (continuous ingest from object storage) | [ClickPipes](/integrations/clickpipes) for S3, GCS, and Azure Blob Storage | Managed ingest from object storage. |
```
Blargian: This might make it sound like we only support S3, GCS, Azure Blob Storage, which is not the case. Let's link to https://clickhouse.com/docs/integrations/clickpipes for a full list of supported sources.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Snowpipe Streaming | [ClickPipes](/integrations/clickpipes) for Kafka, Kinesis, Pub/Sub, MongoDB, MySQL, Postgres, and BigQuery | Managed low-latency streaming and CDC ingest. |
```
Blargian: MongoDB, MySQL, Postgres, BigQuery seem unrelated to streaming. Let's refer users to https://clickhouse.com/docs/integrations/clickpipes#supported-data-sources - streaming sources are marked as such

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Openflow connectors | [ClickPipes](/integrations/clickpipes) and the broader [integrations library](/integrations) | ClickPipes is ClickHouse Cloud's managed connector platform; coverage spans streaming systems, OLTP sources, and object storage. See the integrations library for the current source list. |
+| Kafka connector | [ClickPipes for Kafka](/integrations/clickpipes/kafka), or the [Kafka table engine](/engines/table-engines/integrations/kafka) for self-managed pipelines | Same role; ClickPipes is the managed option. |
+| Snowflake Connector for Postgres / MySQL | [ClickPipes for Postgres](/integrations/clickpipes/postgres), [MySQL](/integrations/clickpipes/mysql) | Managed CDC from OLTP sources. |
```
Blargian: Let's call out Managed Postgres here

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Snowpark (Python / Java / Scala DataFrames) | External Python with [`clickhouse-connect`](/integrations/python) or another [client library](/integrations/python) | No in-database DataFrame runtime; notebook-side libraries cover the same workflow. |
+| Snowflake ML (in-database training; formerly Snowpark ML) | External training and serving (notebooks, Spark, Vertex AI, feature stores) reading from ClickHouse; see [AI/ML in Cloud](/cloud/features/ai-ml) for managed-side features | ClickHouse has no in-database ML. Use it as the analytical store and run training elsewhere. |
+| Cortex LLM functions (`CORTEX.COMPLETE`, `CORTEX.SUMMARIZE`, etc.) | — | No in-SQL LLM surface; call hosted models from the application layer or an orchestrator and write results back to ClickHouse. |
+| Cortex Analyst | [ClickHouse Agents](/cloud/features/ai-ml/agents) in the cloud console | Agents (currently in beta) are conversational: natural-language queries against your data with tool calls and chat workflows. Check the [Agents](/cloud/features/ai-ml/agents) page for the current capability surface. |
```
Blargian: let's avoid "currently in beta" as it will go scale - users can see the status from the badge in the linked section

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Snowsight (web UI: editor, dashboards, monitoring, admin) | ClickHouse Cloud console, which includes [SQL Console](/integrations/sql-clients/sql-console), service management, monitoring, and dashboards | The ClickHouse Cloud console is the equivalent web surface; SQL Console is one component of it, not the whole UI. |
+| Streamlit in Snowflake / Native Apps / Snowpark Container Services | — | ClickHouse has no in-product app-hosting, container, or app-distribution layer. Host Streamlit, container workloads, and packaged apps externally, then query ClickHouse over its native protocol or HTTP. |
+| Notebooks in Snowflake | Jupyter with [`clickhouse-connect`](/integrations/python) or another [client library](/integrations/python) | No in-product notebook environment; notebook-side libraries cover the same workflow. |
```
Blargian: We have a first-class integration with HEX for notebooks

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,201 @@
...
+| Data lineage / Snowflake Horizon Catalog | [`system.*`](/operations/system-tables) tables for metadata; external tools (dbt, DataHub) for lineage and quality | ClickHouse exposes metadata via system tables rather than a managed catalog product. |
+| Database replication / Account replication / Failover Groups (Snowgrid) | In-region high availability via multiple replicas (managed by Cloud); cross-region resiliency via [backups](/cloud/manage/backups/overview); see [Disaster recovery](/cloud/data-resiliency) | ClickHouse Cloud has no automatic cross-region replication, failover, or active-active sync today. Migrating from a Failover Groups deployment means restructuring DR around scheduled [backups](/cloud/manage/backups/configurable-backups) (configurable to every 6 hours, exportable cross-region to a customer-owned bucket) and manual restore into a new service. RPO becomes the backup interval; there's no replicated standby to fail over to. |
```
Blargian: True, but will go stale in the near future re. cross-region replication, failover etc. I think maybe it's better just to link to `/cloud/data-resiliency` as the source of truth.

SOmething like "See article ["data resiliency"]() for more details"

Blargian (changes requested):
@dhtclk structure looks great - left some comments around accuracy of some statements. We will need a few more pairs of eyes on this as well.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+doc_type: 'reference'
+---
+
+The tables below map each Snowflake concept to its ClickHouse equivalent. For function-by-function SQL syntax mapping, see the [SQL translation reference](/migrations/snowflake-translation-reference). For the end-to-end migration walkthrough, see [Migrating from Snowflake to ClickHouse](./02_migration_guide.md).
+
+## Resource hierarchy {#resource-hierarchy}
+
+| Snowflake | ClickHouse | Notes |
+|---|---|---|
+| Organization | [Organization](/cloud/security/console-roles#organization-roles) | Root node of the hierarchy in both. |
```
amychen1776: This is actually not correct. Snowflake's account is actually the Clickhouse equivalent to what we call an org

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Account | [Warehouse](/cloud/reference/warehouses) | Each service scales compute independently; storage is shared at the warehouse level. Tier and billing are set at the organization level, not per warehouse. |
```
amychen1776: Per above - this is also not correct. 
A service in CH is actually for us the equivalent of a database + warehouse in Snowflake.

Warehouses (CH) are how you have multiple warehouses that can read/write to one database (Snow). But please note that it could be multiple databases in Snow. They have true decoupled storage(the data) + compute

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Database | [Database](/sql-reference/statements/create/database) | Logical container for tables. Snowflake uses a Database → Schema → Table hierarchy; ClickHouse flattens this to Database → Table. See [Schemas](#schemas) below. |
```
amychen1776: it's not just tables - it's also views

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+
+:::note[Warehouse terminology]
```
amychen1776, line 20: def let's keep this and link to the warehouses page

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Schemas {#schemas}
+
+A Snowflake schema serves multiple roles and has no single equivalent in ClickHouse.
+
+| Snowflake | ClickHouse | Notes |
+|---|---|---|
+| Namespace partitioning — letting objects with the same name coexist (`analytics.users` vs `marketing.users`) | One [database](/sql-reference/statements/create/database) per Snowflake schema, or fold the schema name into the database (`analytics.public.events` → `analytics_public.events`) | Object references move from three-level (`DB.SCHEMA.TABLE`) to two-level (`DB.TABLE`). |
+| Logical grouping by domain or processing stage (`analytics.raw`, `analytics.staging`, `analytics.marts`) | Separate databases or a consistent naming convention | — |
+| Permission boundary | [SQL grants](/sql-reference/statements/grant) at the database, table, or column level | Database-wide grants cover the schema-level grant footprint; per-table grants are also available for finer-grained control. |
```
amychen1776: what does permission boundary mean?

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Future grants | Database wildcards (`GRANT … ON db.* TO role`) apply to current and future tables | Can't scope future grants to a subset of tables within a database. |
```
amychen1776: funny enough - I'm not quite sure what this is about `can't scope future grants to a subset of tables within a database.`

amychen1776: Is this a CH limitation?

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Roles and access control {#roles-access}
+
+ClickHouse Cloud's access layer splits into [console roles](/cloud/security/console-roles) at `console.clickhouse.cloud` (organization-, service-, and SQL-console-scoped) for org admin, billing, and service management; and SQL [roles and grants](/sql-reference/statements/grant) inside each service for database, table, and column access.
+
+| Snowflake | ClickHouse | Notes |
+|---|---|---|
+| Account-level system roles (`ACCOUNTADMIN`, `SYSADMIN`, `SECURITYADMIN`, `USERADMIN`, `PUBLIC`) | [Organization roles](/cloud/security/console-roles#organization-roles) (Admin, Billing, Org API reader, Member) and [service roles](/cloud/security/console-roles#service-roles) (Service admin, Service reader, Service API admin/reader) in the console; SQL roles inside each service | Org-scoped console roles cover billing, org admin, and user management; service-scoped roles cover service config, scaling, and backups. |
+| Custom account roles | [`CREATE ROLE`](/sql-reference/statements/create/role) in SQL | Same pattern: create a role, grant privileges to it, grant the role to users. |
```
amychen1776: custom account roles are more equivalent to our custom roles on CH Cloud. Create role is more of a db level role which snowflake's roles are completely unified db + the actual cloud account

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Database roles | — | ClickHouse has only one tier of SQL roles, all service-scoped. No equivalent to Snowflake's two-tier account/database role split. For per-user scoped SQL console access, see the callout below. |
```
amychen1776: that's something about service scoped that sounds wrong. It's actually more warehouse scopes if they're using compute-compute separation but not everyone is on that....  But also custom roles also covers both db and cloud

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+:::
+
+## Compute and capacity {#compute-capacity}
+
+| Snowflake | ClickHouse | Notes |
+|---|---|---|
+| Virtual warehouse | Service (one or more replicas) | [Basic](/cloud/manage/cloud-tiers#basic) tier services are single-replica; [Scale and Enterprise](/cloud/manage/cloud-tiers#scale) support multi-replica (2+) deployments for higher SLAs. Queries [parallelize](/optimize/query-parallelism) across replicas. |
+| Warehouse size (XS through 6X-Large) | Vertical [autoscaling](/cloud/features/autoscaling/vertical) bounds | Sizing is configured as min/max memory and CPU bounds rather than discrete t-shirt sizes; setting min = max effectively fixes the size. |
```
amychen1776: >setting min = max effectively fixes the size.

I don't know what this actually means? Like the range is the size?

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Multi-cluster warehouse | Manual [horizontal scaling](/cloud/features/autoscaling/horizontal) | ClickHouse scales replica count rather than cluster count. There's no direct equivalent to Snowflake's auto-scaling policies (`Standard`/`Economy`); horizontal replica count is set manually. |
```
amychen1776: don't we have horizontal autoscaling?

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Query Acceleration Service | No direct equivalent | ClickHouse has no per-query compute booster; scale the service via [vertical autoscaling](/cloud/features/autoscaling/vertical) if queries are consistently large. |
+
+## Billing and pricing model {#billing}
+
+ClickHouse Cloud meters compute as per-minute [compute units (8 GiB RAM, 2 vCPU)](/cloud/manage/billing/overview#how-is-compute-metered) rather than as credits scaled by warehouse size, charges for storage as compressed bytes without Time Travel or Fail-safe overhead, and bills backups as a separate line item rather than bundling them into retention windows. Most Snowflake "serverless compute" features (Snowpipe, Search Optimization, Auto-clustering, materialized view refresh, Cortex) are bundled into service compute on ClickHouse; [ClickPipes](/integrations/clickpipes) is the explicit exception and is [metered separately](/cloud/reference/billing/clickpipes). As in Snowflake, ClickHouse Cloud charges for public internet egress and cross-region data transfer and offers committed-spend discounts. See [ClickHouse Cloud pricing](/cloud/manage/billing/overview) for current rates, tiers, and commitment options.
```
amychen1776:
```suggestion
ClickHouse bills based on dedicated machines, metered in per-minute [compute units (8 GiB RAM, 2 vCPU)](https://github.com/cloud/manage/billing/overview#how-is-compute-metered). Storage pricing is based on compressed bytes, and backups are billed separately. [ClickPipes](/integrations/clickpipes) is [metered separately](/cloud/reference/billing/clickpipes).

Snowflake bills on credits scaled by warehouse size, charges for storage as compressed bytes without Time Travel or Fail-safe overhead, and bills backups as a separate line item rather than bundling them into retention windows. Most Snowflake "serverless compute" features (Snowpipe, Search Optimization, Auto-clustering, materialized view refresh, Cortex) are bundled into service compute on ClickHouse. 

Both charges for public internet egress and cross-region data transfer and offers committed-spend discounts. See [ClickHouse Cloud pricing](/cloud/manage/billing/overview) for current rates, tiers, and commitment options.
```

amychen1776: Just made this a bit more neutral

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Storage and tables {#storage-tables}
+
+In ClickHouse, a table's behavior is set at creation time: the engine (MergeTree family) determines merge and storage semantics, and `ORDER BY` / `PARTITION BY` / `TTL` clauses configure physical layout and retention. Many Snowflake per-feature settings map to a clause in the ClickHouse `CREATE TABLE` statement. Physical schema design also differs between platforms; see the [migration guide](./02_migration_guide.md) for design tradeoffs.
```
amychen1776: I really want to find another word than `behavior`. It just sounds off to me. structure?

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+
+| Snowflake | ClickHouse | Notes |
+|---|---|---|
+| Permanent table | [MergeTree-family table](/engines/table-engines/mergetree-family/mergetree) | Engine choice determines storage and merge behavior; pick by access pattern ([`MergeTree`](/engines/table-engines/mergetree-family/mergetree) for append-mostly facts, [`ReplacingMergeTree`](/engines/table-engines/mergetree-family/replacingmergetree) for upserts, [`AggregatingMergeTree`](/engines/table-engines/mergetree-family/aggregatingmergetree) for pre-aggregations). |
+| Transient table (no Fail-safe) | MergeTree table | ClickHouse has no Fail-safe tier, so the permanent/transient distinction doesn't apply. |
```
amychen1776:
```suggestion
| Transient table (no Fail-safe) | - | ClickHouse has no Fail-safe tier, so the permanent/transient distinction doesn't apply. |
```

We don't have transient table so shouldn't it be this? Since we don't support time travel

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Temporary table (session-scoped) | [`CREATE TEMPORARY TABLE`](/sql-reference/statements/create/table#temporary-tables) | Session-scoped temporary tables exist in both; semantics are similar. |
+| External table | [`s3`](/sql-reference/table-functions/s3) / [`gcs`](/sql-reference/table-functions/gcs) / [`azureBlobStorage`](/sql-reference/table-functions/azureBlobStorage) table functions for direct file access; [Iceberg engine](/engines/table-engines/integrations/iceberg) for open catalogs | Object storage and open-table formats are read directly through these functions and engines. |
+| Stage (internal / external / user / table) | Object storage referenced directly via [`s3`](/sql-reference/table-functions/s3) / [`gcs`](/sql-reference/table-functions/gcs) / [`azureBlobStorage`](/sql-reference/table-functions/azureBlobStorage) table functions; [ClickPipes](/integrations/clickpipes) for managed staging on load | ClickHouse has no stage object: there's no managed internal storage layer for files awaiting load, and no `PUT` / `GET` equivalents for moving files in and out. Read from the bucket directly, or use ClickPipes to coordinate ingest. |
```
amychen1776:
```suggestion
| Stage (internal / external / user / table) | Object storage referenced directly via [`s3`](/sql-reference/table-functions/s3) / [`gcs`](/sql-reference/table-functions/gcs) / [`azureBlobStorage`](/sql-reference/table-functions/azureBlobStorage) table functions; [ClickPipes](/integrations/clickpipes) for managed staging on load | ClickHouse does not support an intermediary staging step. Read from the bucket directly, or use ClickPipes to coordinate ingest. |
```

amychen1776: this sounds nicer to us?

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Iceberg table (managed or unmanaged) | [Iceberg engine](/engines/table-engines/integrations/iceberg) | See the [data lake support matrix](/use-cases/data-lake/support-matrix) for read, write, and storage-backend support. |
+| Snowflake Open Catalog (Polaris) | [Iceberg engine](/engines/table-engines/integrations/iceberg) with REST catalog support | ClickHouse reads from a REST catalog but isn't itself a catalog server. |
```
amychen1776: might be worth calling out that we do not host our own iceberg catalog

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Secure view | [View](/sql-reference/statements/create/view) with `SQL SECURITY DEFINER` | `SQL SECURITY DEFINER` delegates privileges (the view runs as its owner) but isn't a full Secure View: the definition stays readable via `SHOW CREATE` by anyone who can query it. See [CREATE VIEW](/sql-reference/statements/create/view) for the `DEFINER` / `INVOKER` / `NONE` modes. |
+| Row access policy | [Row policy](/sql-reference/statements/create/row-policy) — a `WHERE`-style expression evaluated per user | Row policies apply transparently to every query against the table. |
+| Sequence | [`generateSerialID`](/sql-reference/functions/other-functions#generateSerialID) for a Keeper-backed sequential counter; [`generateSnowflakeID`](/sql-reference/functions/uuid-functions#generateSnowflakeID) or [`generateUUIDv7`](/sql-reference/functions/uuid-functions#generateUUIDv7) for distributed unique IDs | `generateSerialID` is the closest match to an auto-incrementing sequence: a named, monotonic counter coordinated through ClickHouse Keeper. The UUID functions suit high-throughput unique IDs that don't need a shared counter. |
+
+:::note[Time Travel and backups]
```
amychen1776: have we thought about just having this part and not including these features in the table? to lessen up the mentions?

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+non-trivial to change later.
+:::
+
+## Query model and performance {#query-model}
+
+Query acceleration in ClickHouse comes from three layers: primary-key ordering (a sparse index over the on-disk sort order), secondary indexes on non-key columns, and materialized views.
+
+| Snowflake | ClickHouse | Notes |
+|---|---|---|
+| Primary key (advisory) | Primary key — drives the on-disk sort order and the [sparse primary index](/guides/best-practices/sparse-primary-indexes) | Where Snowflake's PK is advisory only, ClickHouse's PK is load-bearing — it determines physical layout and is used to prune granules, avoid re-sorts, and short-circuit `LIMIT`. Neither system enforces uniqueness. |
```
amychen1776: we should explicitly call out the fact that our primary key does not have to be unique. That's like an industry standard (that PKs are unique)

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Foreign key (advisory) | Wide tables or [dictionaries](/dictionary) for lookups | ClickHouse doesn't accept foreign-key declarations even as advisory hints. |
```
amychen1776: Are we talking about foreign key constraints or...? I'm confused by this because foreign keys to me are just the join key

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Cortex Search / Snowflake Cortex Search | [Full-text index](/engines/table-engines/mergetree-family/textindexes) | Token index over string columns for in-database search. |
+| `VECTOR` data type and vector search | [`Array(Float32)`](/sql-reference/data-types/array) or [`Array(BFloat16)`](/sql-reference/data-types/float#bfloat16) with a [vector ANN index](/engines/table-engines/mergetree-family/annindexes); or [`QBit`](/sql-reference/data-types/qbit) for tunable-precision search | ClickHouse has no dedicated `VECTOR` type. Embeddings store as `Array(Float32)`, or `Array(BFloat16)` to halve storage, with an ANN index accelerating approximate nearest-neighbor lookups. `QBit` keeps full precision while letting you trade bits for speed at query time. |
+| Materialized view | [Incremental MV](/materialized-view/incremental-materialized-view) — updates on each insert into a base table | Source-shape rules differ; review both before porting an existing MV. Cost is paid at insert time in ClickHouse. |
```
amychen1776, line 129: fun fact - Snowflake views are extremely limited and don't even support joins :)

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Dynamic data masking (function-based) | [`CREATE MASKING POLICY`](/sql-reference/statements/create/masking-policy) (ClickHouse Cloud); or [views](/sql-reference/statements/create/view) and [row policies](/sql-reference/statements/create/row-policy). See [data masking patterns](/cloud/guides/data-masking) | `CREATE MASKING POLICY` is a direct equivalent: function-based column masking applied at query time per role, without changing stored data. |
+| Network policies (IP allowlist) | IP allowlists and [private connectivity](/cloud/security/connectivity/private-networking) — PrivateLink (AWS, Azure) and Private Service Connect (GCP) for ingress restriction | Private connectivity is available across the three major clouds. |
+| Tri-Secret Secure (customer-managed keys) | [CMEK](/cloud/security/cmek) on the service | Supports key rotation and revocation. See the CMEK page for the current list of supported cloud providers. |
+| Object tagging (governance metadata) | — | ClickHouse exposes metadata via `system.*` tables rather than user-defined tags. |
+| Data classification (sensitive-data detection) | — | Not a managed feature; external tools (e.g. DataHub) cover this layer. |
```
amychen1776, line 159: We do support tagging but it's definitely not to level of Snowflake

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
morsapaes: Should be "database objects", or just "objects", as it includes any object that can be created in SQL (_e.g._ tables, views, materialized views, named collections, indexes), which are all bound to a namespace. The more correct way to represent the namespace hierarchies would be `database.object` (ClickHouse) and `database.schema.object` (Snowflake).

morsapaes: And these are not equivalent. The correct equivalent concepts would be `Schema` in Snowflake, `Database` in ClickHouse.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+
+:::note[Warehouse terminology]
+A **ClickHouse warehouse** is a grouping of services that share
+storage and scale compute independently, not a compute cluster as
+in Snowflake.
+:::
+
+## Schemas {#schemas}
+
+A Snowflake schema serves multiple roles and has no single equivalent in ClickHouse.
```
morsapaes, line 26: A schema in Snowflake is technically equivalent to a database in ClickHouse.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Account | [Warehouse](/cloud/reference/warehouses) | Each service scales compute independently; storage is shared at the warehouse level. Tier and billing are set at the organization level, not per warehouse. |
+| Database | [Database](/sql-reference/statements/create/database) | Logical container for tables. Snowflake uses a Database → Schema → Table hierarchy; ClickHouse flattens this to Database → Table. See [Schemas](#schemas) below. |
+
+:::note[Warehouse terminology]
+A **ClickHouse warehouse** is a grouping of services that share
+storage and scale compute independently, not a compute cluster as
+in Snowflake.
+:::
+
+## Schemas {#schemas}
```
morsapaes, line 24: Not sure this section makes sense as is. It should probably be a subsection of the one above that explains how to map the Snowflake namespace hierarchy to the more restrictive hierarchy we use in ClickHouse. There are know issues when users migrate, _e.g._ with integrations like dbt.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
morsapaes: Related to [this comment](https://github.com/ClickHouse/clickhouse-docs/pull/6244/changes#r3389550636), I think this section needs to be folded into the previous one and use a different format. The namespace hierarchies are fundamentally different, so it doesn't make sense to compare apples to oranges.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Future grants | Database wildcards (`GRANT … ON db.* TO role`) apply to current and future tables | Can't scope future grants to a subset of tables within a database. |
+| Schema `OWNERSHIP` and `MANAGED ACCESS` | — | ClickHouse has no object-ownership model, so grants are always explicit. |
+| Cloning unit (`CREATE SCHEMA … CLONE`) | Per-table [`CREATE TABLE ... CLONE AS`](/sql-reference/statements/create/table#with-a-schema-and-data-cloned-from-another-table), not per-schema; see the [Storage and tables](#storage-tables) section | No single-statement clone of a whole schema or database. Clone each table individually. |
+| Time Travel and replication boundary | Service level — point-in-time recovery via [backups](/cloud/manage/backups/overview); replication via managed replicas, see [Operations and ecosystem](#operations) | No per-schema boundary: recovery and replication are scoped per service. TTL controls data expiry, not point-in-time recovery. See the [Data retention](#storage-tables) row. |
+| Tagging and classification scope | Apply at the table or column level | No intermediate namespace inherits down. |
+
+## Roles and access control {#roles-access}
```
morsapaes, line 39:
```suggestion
## Access control {#access-control}
```

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Role hierarchy (`GRANT ROLE … TO ROLE …`) | [`GRANT role1 TO role2`](/sql-reference/statements/grant) | — |
+| Privilege grants on objects (`GRANT … ON … TO ROLE …`) | [`GRANT … ON db.table TO role`](/sql-reference/statements/grant) | — |
+| Object ownership and ownership transfer | — | Access in ClickHouse is controlled entirely through explicit grants. Snowflake patterns that rely on owners delegating access need to be rebuilt as explicit role-based grants. |
+| `USE ROLE` | [`SET ROLE`](/sql-reference/statements/set-role) | — |
+
+:::note[Per-user SQL console access]
+ClickHouse has no separate database-role tier, but ClickHouse Cloud can still grant per-user, scoped access to the SQL console. Create a role matching the `sql-console-role:<email>` naming convention and grant it the privileges that user should have; the console assigns it in place of the default `sql_console_admin` / `sql_console_read_only` roles. These are ordinary service-scoped SQL roles — the namespace is just a mapping convention. See [granular access control](/cloud/security/common-access-management-queries#granular-access-control).
+:::
```
morsapaes, line 55: This note is confusing. What is a "database-role" tier? It feels like we're introducing a lot of non-standard terminology in this section, in general.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
morsapaes: It's in Private Preview, but for this purpose we could just say "Horizontal autoscaling" and mention it's based on work load.

morsapaes: @dominic-wright, see [this #shipped post](https://clickhouse-inc.slack.com/archives/C0AEE7J8U8G/p1779810729558109).

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
```diff
@@ -0,0 +1,202 @@
...
+| Warehouse size (XS through 6X-Large) | Vertical [autoscaling](/cloud/features/autoscaling/vertical) bounds | Sizing is configured as min/max memory and CPU bounds rather than discrete t-shirt sizes; setting min = max effectively fixes the size. |
+| Multi-cluster warehouse | Manual [horizontal scaling](/cloud/features/autoscaling/horizontal) | ClickHouse scales replica count rather than cluster count. There's no direct equivalent to Snowflake's auto-scaling policies (`Standard`/`Economy`); horizontal replica count is set manually. |
+| Auto-suspend / auto-resume | Service [idling](/cloud/features/autoscaling/idling) | Compute stops when there's no work, restarts on the next query. |
+| Resource monitors (credit-quota spend caps) | [Workloads](/operations/workload-scheduling) for runtime scheduling; per-query limits (memory, threads, execution time) | ClickHouse workloads cover runtime resource scheduling but not spend caps; there's no primitive that suspends a service on hitting a credit threshold. |
```
morsapaes, line 65: We do have billing thresholds and threshold-based notifications, might be worth mentioning. These are only informational, though; we don't cap or restrict usage.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
morsapaes: Can we just follow the current documentation? This paragraph on billing is pretty convoluted. [Here](https://clickhouse.com/docs/cloud/manage/billing/overview), we simply say:

> ClickHouse Cloud bills based on the usage of compute, storage, [data transfer](https://clickhouse.com/docs/cloud/manage/network-data-transfer) (egress over the internet and cross-region), and [ClickPipes](https://clickhouse.com/docs/integrations/clickpipes).

More direct to understand. Backups are lumped into storage costs, I don't think we need to (or should) mention them upfront.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
dhtclk: Good catch this was a miss on my part.  Wildcard grants (GRANT … ON db.* TO role) automatically cover tables created later. I'll update.

docs/cloud/onboard/02_migrate/01_migration_guides/04_snowflake/04_equivalent-concepts.md
morsapaes: "Merge" is a ClickHouse-specific concept, +1 on simplifying this sentence. It feels like we've described this a million times, we should be able to reuse existing descriptions:

In ClickHouse, you define storage and data layout upfront at table creation time. A `CREATE TABLE` statement specifies not only the columns and data types, but also the table engine and the sorting and indexing strategy through an `ORDER BY` clause. The `ORDER BY` clause is equivalent to a Snowflake clustering key: it defines how data is sorted on disk and indexed. In ClickHouse, unlike Snowflake, you don't incur additional background costs for maintaining the sort order once the table is created. This gives you direct control over query performance and storage costs.

Other clauses like `PARTITION BY` or `TTL` are available for partitioning, retention, and other data management strategies, as needed. Many of the settings you configure per-feature in Snowflake map to these clauses in a single `CREATE TABLE` statement. See the migration guide for design tradeoffs.

morsapaes (commented):
Didn't manage to review the whole PR yet, but adding some suggestions for what I was able to review this week.

PR #6373, opened 2026-06-11
Add BYOC shared-responsibility page

Blargian:
## Checklist
- [ ] URL changes should add a redirect to the old URL via https://github.com/ClickHouse/clickhouse-docs/blob/main/vercel.json


BYOC Security Shared Responsibilities.pdf
dhtclk, line 1: Was this intentional?

Blargian: Unintentional

PR #6437, opened 2026-06-23
Add Apache Airflow provider integration page

BentsiLeviav:
## Checklist
- [ ] URL changes should add a redirect to the old URL via https://github.com/ClickHouse/clickhouse-docs/blob/main/vercel.json


chernser (approved):

docs/integrations/data-ingestion/etl-tools/airflow-and-clickhouse.md
```diff
@@ -0,0 +1,212 @@
...
+| `client_name`          | Client Name                  | (empty) | A label appended to the Airflow version identifier in the ClickHouse `User-Agent` and the `client_name` column of [`system.query_log`](/operations/system-tables/query_log). |
+| `session_settings`     | Session Settings (JSON)      | (empty) | [ClickHouse session settings](/operations/settings/settings) applied to every query on the connection, for example `{"max_execution_time": 300, "max_threads": 8}`. |
+| `client_kwargs`        | Client kwargs (JSON)         | (empty) | Additional keyword arguments forwarded to `clickhouse_connect.get_client()`, for example an `http_proxy`. |
+
+### Define a connection without the UI {#define-a-connection-without-the-ui}
+
+Set the connection through an environment variable. The URI form covers host, credentials, and database:
+
+```bash
+export AIRFLOW_CONN_CLICKHOUSE_DEFAULT='clickhouse://default:password@localhost:8123/my_database'
```
koletzilla, line 72: I'm not 100% sure about this, but I think we need to use `clickhousedb` as the driver name here.

I think `clickhouse` also works but that is also used by the module `clickhouse-sqlalchemy`, so it may cause problems if both modules are installed. 

Checking the code of `clickhouse-connect` I can see both strings may be used, so to be honest I'm not sure which should we add here 🤔  CC @joe-clickhouse 
- `clickhousedb`, which looks like the "official" one as it's defined at setup time: https://github.com/ClickHouse/clickhouse-connect/blob/main/setup.py#L83-L84
- `clickhouse`, which is defined only if this class is imported, which made me thing it's not available all the time: https://github.com/ClickHouse/clickhouse-connect/blob/main/clickhouse_connect/cc_sqlalchemy/__init__.py#L13-L14

dhtclk (approved):

koletzilla (approved):

docs/integrations/data-ingestion/etl-tools/airflow-and-clickhouse.md
joe-clickhouse: @koletzilla this is true, however these are two different layers. Here `clickhouse://` is the Airflow connection scheme, which just has to match the provider's `conn_type` of `clickhouse`, while `clickhousedb` is the SQLAlchemy dialect name. The provider already uses `clickhousedb://` internally when it builds a SQLAlchemy engine, so the doc is correct as written and needs no change.

koletzilla: oh that's fantastic, thanks!

PR #6508, opened 2026-07-07
chdb: rewrite JupySQL guide on the NYC taxi dataset

ShawnChen-Sirius:
## Why

The chDB JupySQL guide can no longer be followed end to end:

- Its example data came from the `JeffSackmann/tennis_atp` GitHub repo, which has been **removed** — every download step in the guide now 404s.
- The JupySQL link pointed at `jupysql.ploomber.io`, whose domain no longer resolves (the project itself is alive on GitHub/PyPI).
- The frontmatter description read “How to install chDB for Bun”, copied from another page.

## What

Rewritten around the NYC taxi dataset already hosted for these docs (`datasets-documentation` S3 bucket): three TSV trip files (~3M rows) plus the taxi zone lookup CSV. The structure and teaching points are unchanged from the original — glob queries over raw files, schema cleanup during import (String→Float64 casts and a borocode→borough `transform`), a two-table join, saved queries as CTEs, parametrized queries, and a histogram.

The outdated tennis histogram image import is dropped; the JupySQL link now points at the project's GitHub repository.

## Verification

Every query and output table in the guide was executed against **chdb 4.1.8** on the real downloaded files (3,000,317 trips imported; join and tip results are pasted from actual runs).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


docs/chdb/guides/jupysql.md
```diff
@@ -95,317 +91,269 @@ Next, we'll display the display limit so that results of queries won't be trunca
...
++---------------+------------+---------+----------------+
+|    borough    | zone_count |  trips  | trips_per_zone |
++---------------+------------+---------+----------------+
+|   Manhattan   |     69     | 2713990 |    39333.0     |
+|     Queens    |     69     |  187737 |     2721.0     |
+|    Brooklyn   |     61     |  52445  |     860.0      |
+|    Unknown    |     2      |  43802  |    21901.0     |
+|     Bronx     |     43     |   2300  |      53.0      |
+| Staten Island |     20     |    43   |      2.0       |
++---------------+------------+---------+----------------+
```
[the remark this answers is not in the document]
ShawnChen-Sirius: The lookup file does contain an `Unknown` borough: rows 264 (`NV`) and 265 (`NA`) of `taxi_zone_lookup.csv` both carry `Borough = 'Unknown'` — that's exactly the `zone_count = 2` in the sample output:

```
LocationID 264 │ Unknown │ NV │ N/A
LocationID 265 │ Unknown │ NA │ N/A     (265 rows total)
```

The inner join therefore matches those 43,802 trips as well, and the six rows shown sum to the full 3,000,317. The output table was pasted from a real run against the downloaded files (chdb 4.1.8).

docs/chdb/guides/jupysql.md
```diff
@@ -2,15 +2,12 @@
...
 ---
 
-import Image from '@theme/IdealImage';
-import PlayersPerRank from '@site/static/images/chdb/guides/players_per_rank.png';
-
-[JupySQL](https://jupysql.ploomber.io/en/latest/quick-start.html) is a Python library that lets you run SQL in Jupyter notebooks and the IPython shell.
+[JupySQL](https://github.com/ploomber/jupysql) is a Python library that lets you run SQL in Jupyter notebooks and the IPython shell.
 In this guide, we're going to learn how to query data using chDB and JupySQL.
 
 <div class='vimeo-container'>
```
dhtclk: We may want to add an admonition note below this video indicating that it is referencing a stale dataset that no longer exists. I think the video content is still good and legitimate, but I think without a note the user experience is a little confusing if following the guide.

Perhaps something like this:

```
:::note
The video above follows an earlier version of this guide built on the tennis ATP dataset. The JupySQL and chDB workflow it shows is still accurate, but the written steps below use the New York City taxi dataset instead.
:::
```

dhtclk (commented):
Looks good, just one small suggestion before approval.

docs/chdb/guides/jupysql.md
Blargian: What if we removed the video here and linked it at the bottom rather? Something like:

```
## See also {#see-also}
- [Intro to JupySQL with chDB](https://www.youtube.com/watch?v=2wjl3OijCto&source_ve_path=MTc4NDI0)
```

dhtclk: Ah that also works, probably the simpler solution! Your call on which you'd prefer @ShawnChen-Sirius

dhtclk (approved):

PR #6533, opened 2026-07-15
Docs: Improve API key setup guide

Blargian:
## Summary

- convert the API key setup instructions to the `VerticalStepper` component
- add a dedicated step for configuring allowed IP addresses
- explain how API key-level IP allowlists apply to Query API endpoints and link to the detailed guidance
- reuse the existing allowed-IP configuration screenshot
- fix a nearby typo in the API key management step

## Why

The existing guide presents the setup flow as a plain numbered list and does not include the allowed-IP configuration available when creating an API key. Making each action an explicit step improves scanability and keeps the security guidance visible during setup.

## User impact

Readers get a clearer end-to-end API key creation flow and can restrict API key access to specific IP addresses or CIDR ranges without having to discover the setting elsewhere.

## Validation

- `git diff --check`
- `markdownlint-cli2 --config ./scripts/.markdownlint-cli2.yaml docs/cloud/features/06_admin_features/api/openapi.md` (0 errors)


dhtclk (approved):
LGTM. As a side note, this looks to be a good candidate for future templating.

